PR #12, opened 2026-04-15
perf(parsing): eliminate double parsing and optimize tree-sitter trav…

lemon07r:
…ersal

- Add unified parse_file() that parses once and reuses the AST for symbol extraction, chunking, and reference collection
- Replace per-level cursor allocation with single TreeCursor traversal in collect_symbols and collect_calls
- Use child_by_field_name() for R binary_operator (rhs field) instead of iterating all children
- Add extract_references_with_symbols() to accept pre-computed symbols, avoiding redundant extract_symbols() calls
- Update pipeline.rs and update.rs to use parse_file() for non-RST languages (RST still needs separate parsing due to preprocessing)

---
## Summary by cubic
Parse each file once with tree-sitter and reuse the AST for chunking and call-site references, removing duplicate parsing and extra AST walks. RST stays a special case due to preprocessing; the update flow now preserves refs even if chunking fails.

- **Performance**
  - Added parse_file() to build chunks and references in one pass; returns (chunks, refs).
  - Updated pipeline and update paths to use parse_file(); RST still preprocesses (refs from raw, chunks from preprocessed).
  - Switched symbol and call traversal to a single TreeCursor to reduce allocations and walks.
  - Tweaks: use child_by_field_name("rhs") for R binary_operator; added extract_references_with_symbols() to reuse precomputed symbols.

- **Bug Fixes**
  - Preserved symbol depth with single-cursor traversal to keep correct nesting.
  - In updates, always store RST references from raw source even if preprocessing/chunking fails.

<sup>Written for commit 0fee201538b170ed1400c6ce371edf9bc446d113. Summary will update on new commits.</sup>

### DIFF
--- a/crates/vera-core/src/indexing/pipeline.rs
+++ b/crates/vera-core/src/indexing/pipeline.rs
@@ -317,10 +317,11 @@ fn parse_discovered_files_parallel(
                         Language::from_extension(ext)
                     });
 
-                let refs = parsing::parse_and_extract_references(&source, language);
-
-                let normalized_source = if language == Language::Rst {
-                    match parsing::sphinx::preprocess_rst(
+                // RST files need preprocessing before chunking, but refs
+                // come from the raw source, so they can't share a single parse.
+                let parse_result = if language == Language::Rst {
+                    let refs = parsing::parse_and_extract_references(&source, language);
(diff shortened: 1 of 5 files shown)


lemon07r:
@codex[agent]+gpt-5.4 review

PR #16, opened 2026-04-20
fix(local): align CoreML reranker and search filters

lemon07r:
## Summary
- switch CoreML local reranker downloads and probes from the quantized Jina export to the FP16 export, while keeping other backends on the existing quantized path
- align the CLI and skill-facing search surface, then restore the MCP and regex filter paths so advertised `lang`/`path`/`symbol_type` options are actually enforced
- update docs and regression tests around CoreML asset repair and regex/MCP filtering behavior

## Testing
- cargo test --workspace
- cargo clippy --workspace -- -D warnings


lemon07r:
@codex[agent]+gpt-5.4 review

PR #24, opened 2026-06-09
fix(search): treat bare directory pattern as prefix match in path glob (Closes #23)

### LINKED ISSUE #23: Support vera search something under special dir
I can not use `vera search something --path src/app` to search something under some dir
is it a bug?

This test can not pass, 
```
#[test]
fn filter_by_path_plain_directory_prefix() {
    let filters = SearchFilters {
        path_glob: Some("app/src".to_string()),
        ..Default::default()
    };
    let inside = make_test_result("app/src/foo.ts", Language::TypeScript, None, None);
    let deep = make_test_result("app/src/bar/baz.rs", Language::Rust, None, None);
    let sibling = make_test_result("app/srcs/foo.ts", Language::TypeScript, None, None);
    let other = make_test_result("other/src/foo.ts", Language::TypeScript, None, None);
    assert!(filters.matches(&inside));
    assert!(filters.matches(&deep));
    assert!(!filters.matches(&sibling));
    assert!(!filters.matches(&other));
}
```

flier268:
--path app/src did not match files under that directory because
  glob_match_recursive returned false when the pattern was exhausted
  but the path still had a trailing segment (e.g. "/foo.ts").

  Change the empty-pattern base case to also accept text that starts
  with '/', giving "app/src" the same effect as "app/src/**".
 (Closes #23)

---
## Summary by cubic
Fix path-glob matching so bare directory paths are treated as a prefix without loosening wildcard rules. `--path app/src` now matches files under that directory (same as `app/src/**`), and `src/*` remains single-segment.

- **Bug Fixes**
  - Added a directory-prefix fallback in `glob_matches` only for patterns with no wildcards.
  - Added tests for plain directory prefix behavior and for ensuring `src/*` does not cross directory boundaries.

<sup>Written for commit 682ce28d901e69c8fc1a56270bb8c48c097af582. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/lemon07r/Vera/pull/24?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/vera-core/src/types.rs
+++ b/crates/vera-core/src/types.rs
@@ -145,7 +145,19 @@ fn glob_matches(pattern: &str, path: &str) -> bool {
     let pattern = pattern.replace('\\', "/");
     let path = path.replace('\\', "/");
 
-    glob_match_recursive(&pattern, &path)
+    if glob_match_recursive(&pattern, &path) {
+        return true;
+    }
+
+    // Directory-prefix fallback: a bare pattern with no wildcards (e.g. "app/src")
+    // should match any file beneath that directory ("app/src/foo.ts").
+    // This is intentionally kept out of the recursive matcher so that wildcard
+    // handlers like `*` retain strict single-segment semantics.
+    if !pattern.contains(['*', '?', '[']) {
+        return path.starts_with(&format!("{pattern}/"));
+    }
+
+    false
 }
 
 /// Recursive glob matching helper.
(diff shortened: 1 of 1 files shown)


lemon07r:
Closing in favor of #28, which carries the complete version of this fix (directory-prefix fallback kept out of the recursive matcher so wildcard semantics are preserved, plus normalization and fuller tests). Thank you for the original report and PR, @flier268 — the issue was real and your report got it fixed.

PR #22, opened 2026-06-08
feat(serve): add vera serve command with pure inference API

flier268:
## Summary by cubic
Adds a new `vera serve` command and the `vera-serve` crate to run an HTTP inference server for remote embeddings and reranking, with per-request model loading and an optional idle-timeout cache to reduce cold starts. Client disconnects cancel in-progress work to avoid wasted GPU/CPU cycles.

- **New Features**
  - New `vera serve` CLI in `vera-cli` with `--host`, `--port`, `--api-key`, local backend flags (or `--api`), and `--idle-timeout` (0=reload per request, -1=keep loaded, N=evict after N seconds).
  - New `vera-serve` crate using `axum`; pure compute only (no `--path`, no index). Providers load per request or from the idle cache; background eviction drops idle models.
  - Endpoints: POST `/v1/embeddings` (OpenAI-compatible), POST `/v1/rerank` (Cohere/Jina-compatible, gated by reranker availability), GET `/v1/health`.
  - Optional Bearer auth via `--api-key` or `VERA_SERVE_KEY` with constant-time comparison.
  - `/v1/embeddings` accepts a single string or an array; rerank sorting is NaN-safe.
  - Client disconnects cancel embeddings/rerank between sub-batches and return HTTP 499.

- **Bug Fixes**
  - Prevent empty-key auth bypass (`VERA_SERVE_KEY=""` or `--api-key ""` now disable auth rather than pass).
  - Redact internal error details from 500 responses; full errors are logged server-side.
  - Fix cancellation gaps: add checks for CPU-mode embeddings and route non-cancellable rerank calls through the cancellable path for consistency.

<sup>Written for commit de7b8ed3be188f6bb82c6fdf1aa3eb44828c027f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/lemon07r/Vera/pull/22?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #32, opened 2026-08-16
v1.0.0 release

### LINKED ISSUE #19: CoreML reranker falls back to CPU on Apple Silicon
## Summary

When using `--onnx-jina-coreml`, the **embedding model** runs on the Apple GPU via CoreML EP, but the **reranker model** (`jina-reranker-v2-base-multilingual`) falls back entirely to CPU. `vera doctor --probe` reports all probes passing, giving no indication that the reranker is CPU-only.

The reranker registers CoreML EP successfully, but ONNX Runtime's CoreML EP partitions **0 out of 667 nodes** to CoreML because the `model_fp16.onnx` export stores all weights and intermediate tensors as `float16` (ONNX type 10), which CoreML EP does not support as input dtype.

## Environment

- M1 Pro, 32 GB, macOS Sequoia
- Vera v0.12.13
- ONNX Runtime: 1.24.4 (coreml build)

## Steps to Reproduce

```bash
vera setup --onnx-jina-coreml
vera config set retrieval.reranking_enabled true
time vera search --deep --compact -n 1 --intent "test" "test query"
```

## Expected Behavior

Reranker inference uses Apple GPU. Search with reranking completes in a few seconds.

## Actual Behavior

Reranker runs entirely on CPU, saturating all cores. A search reranking 141 candidates takes ~15–43 seconds (vs ~0.8s with reranking disabled). CPU usage spikes to ~1000% (all 10 cores).

```bash
# With reranking
real    0m37.023s
user    4m59.552s    # ← 8x real time = all CPU cores maxed
sys     0m4.773s

# Without reranking
real    0m0.826s
user    0m0.837s
sys     0m0.278s
```

## Evidence

I enabled ORT verbose logging (`log_severity_level=0`) against the reranker model with Python `onnxruntime` to get partition diagnostics:

```
CoreMLExecutionProvider::GetCapability, number of partitions supported by CoreML: 0
number of nodes in the graph: 667
number of nodes supported by CoreML: 0
```

Every single node is rejected because CoreML EP doesn't support `float16` tensor inputs:

```
[Add]       Input type: [10] is not currently supported
[ReduceMean] Input type: [10] is not currently supported
[MatMul]    Input type: [10] is not currently supported
[Sqrt]      Input type: [10] is not currently supported
...
```

ONNX type `10` = `float16`. All 263 weight/bias initializers and 606 intermediate value_info entries in `model_fp16.onnx` are `float16`.

## Root Cause

`local_models.rs:41` selects `model_fp16.onnx` for CoreML because the quantized export has `DynamicQuantizeLinear`/`MatMulInteger` ops that CoreML EP also doesn't support:

```rust
/// CoreML cannot execute the quantized reranker export because it contains
/// DynamicQuantizeLinear/MatMulInteger ops that the CoreML EP does not support.
const RERANKER_ONNX_COREML_FILE: &str = "onnx/model_fp16.onnx";
```

The fp16 export was chosen as the workaround, but CoreML EP doesn't support `float16` dtype inputs either. Neither ONNX export works with CoreML EP.

By contrast, the **embedding model** (`jina-embeddings-v5-text-nano-retrieval`) works on CoreML because its graph has fp32 intermediate tensors even though weights are stored as fp16 — ORT's `CastFloat16Transformer` optimizer can upcast it during session creation.

## Secondary Issue: `vera doctor` Doesn't Detect This

`vera doctor --probe` reports:

```
ok    probe-provider-registration registered coreml
ok    probe-provider-confirmation session init and tiny inference succeeded;
      active GPU execution still cannot be confirmed via the ONNX Runtime Rust API
```

The probe only verifies that CoreML EP *registers* and that a tiny inference *completes*. Since ORT silently falls back to CPU when CoreML EP can't partition any nodes, the probe passes while the reranker is actually CPU-only. The `probe-provider-confirmation` message acknowledges this limitation but doesn't surface it as a warning.

**Suggestion:** During `vera doctor`, log the partition result (supported nodes / total nodes) when possible, or at minimum, add a performance warning when the reranker probe takes >N milliseconds on a CoreML backend (CPU fallback is orders of magnitude slower than GPU).

## Possible Fixes

1. **Request a fp32 ONNX export from Jina** — A `model_fp32.onnx` or `model.onnx` with all `float32` tensors would allow CoreML EP to partition nodes. I verified with Python ORT that converting initializers + value_info + ConstantOfShape attributes from fp16→fp32 results in CoreML EP accepting 425/667 nodes (63%). However, ORT's `SimplifiedLayerNormFusion` optimizer then crashes on the type-converted model, so a proper export from PyTorch is needed rather than a post-hoc conversion.

2. **Export fp32 from PyTorch at download time** — When `vera setup --onnx-jina-coreml` runs, use `optimum` or `transformers` to export `jina-reranker-v2-base-multilingual` to ONNX fp32 instead of downloading the pre-built fp16 file.

### LINKED ISSUE #20: `--intent` flag breaks BM25 search with "Field does not exist: 'intent'" error
## Summary

When using `--intent`, the formatted query string `"intent: <text> | <query>"` is passed directly to Tantivy's `QueryParser`, which interprets `intent:` as a Lucene-style field query. Since there is no BM25 field named `intent`, the query fails and BM25 search falls back entirely to vector-only results.

## Steps to Reproduce

```bash
vera search --deep --compact -n 1 --intent "test" "test query"
```

## Expected Behavior

BM25 and vector search both contribute to hybrid retrieval.

## Actual Behavior

```
WARN vera_core::retrieval::hybrid: BM25 search failed, falling back to vector-only results
  error=BM25 search failed for query: intent: test | test query: failed to parse BM25 query: ...
  Field does not exist: 'intent'
```

Every search with `--intent` loses BM25 entirely.

## Root Cause

`search.rs:185` formats the query:

```rust
fn apply_intent(query: &str, intent: Option<&str>) -> String {
    match intent {
        Some(intent) => format!("intent: {intent} | {query}"),
        None => query.to_string(),
    }
}
```

This produces strings like `"intent: test | test query"`. The Tantivy `QueryParser` in `storage/bm25.rs` interprets `intent:` as a field-qualified query (like `title:foo` in Lucene syntax). The BM25 schema has fields `content`, `symbol_name`, `filename`, `path_tokens`, `language` — but no `intent` field.

The `|` pipe character may also interfere with Tantivy's query parser.

## Suggested Fix

The intent prefix should only be prepended for the embedding/vector query (where it provides semantic context) and should be stripped or reformatted before passing to BM25 search. For example:

- BM25 query: use only the raw user query (the second argument)
- Vector query: include the intent prefix for richer semantic matching
- Or: escape/quote the intent text so Tantivy treats it as literal text rather than a field query

### LINKED ISSUE #21: OpenAI embedding API token-limit errors are fatal instead of auto-truncated
## Summary

`embed_batch_resilient` has a truncation-retry mechanism that handles oversized embedding inputs, but it doesn't recognize OpenAI's token-limit error format. Indexing fails with a fatal error instead of automatically truncating and retrying.

## Environment

- Vera: v0.12.13
- Embedding provider: OpenAI `text-embedding-3-small` (8192 token limit)
- M1 Pro, 32 GB, macOS Sequoia

## Reproduction

1. Configure Vera with OpenAI embeddings (API mode):
   ```
   EMBEDDING_MODEL_BASE_URL=https://api.openai.com/v1
   EMBEDDING_MODEL_ID=text-embedding-3-small
   ```
2. Index a repo containing large source files that produce chunks exceeding ~8192 tokens
3. Indexing fails:

```
Error: indexing failed: embedding generation failed: embedding API error (status 400): {
  "error": {
    "message": "Invalid 'input[3]': maximum input length is 8192 tokens.",
    "type": "invalid_request_error",
    "param": null,
    "code": null
  }
}
```

## Root Cause

`context_size_info()` in `crates/vera-core/src/embedding/provider.rs` (~line 604) checks the error message for these patterns:

- `"context size"`
- `"exceed_context_size_error"`
- `"\"n_ctx\""`
- `"too large to process"`
- `"max allowed tokens per submitted batch"`

OpenAI returns `"maximum input length is 8192 tokens"` – none of the above patterns match. The guard returns `None`, the error is treated as non-recoverable, and `embed_batch_resilient` propagates it as fatal instead of triggering its truncation-retry path.

## Suggested Fix

Add `"maximum input length"` to the match list in `context_size_info()`:

```rust
if !lower.contains("context size")
    && !lower.contains("exceed_context_size_error")
    && !lower.contains("\"n_ctx\"")
    && !lower.contains("too large to process")
    && !lower.contains("max allowed tokens per submitted batch")
    && !lower.contains("maximum input length")  // OpenAI
{
    return None;
}
```

And add a regex to extract the token limit from OpenAI's format:

```rust
static OPENAI_MAX_LENGTH_RE: OnceLock<Regex> = OnceLock::new();
// ...
let openai_re = OPENAI_MAX_LENGTH_RE
    .get_or_init(|| Regex::new(r"maximum input length is (\d+) tokens").unwrap());
```

Include `openai_re` in the existing `max_tokens` extraction chain so the truncation logic knows the actual limit.

## Adjacent issue: error output causes display glitches

Just before the error message is printed, the `◐ Parsed into…` and `◐ Generating embeddings` display is repeated, possibly due to newline characters leaking into the output (see [screenshot](https://ss.strco.de/SCR-20260521-j91p.png))

### LINKED ISSUE #23: Support vera search something under special dir

lemon07r:
## v1.0.0 Release

Everything on the `feat/v1-prep` integration branch since master (78 commits). Closes #19, closes #20, closes #21, closes #23.

### Highlights

**Retrieval quality (full 1,251-task Semble suite, `vera-cuda` lane)**

| Variant | nDCG@10 | Recall@1 | Recall@5 |
|---|---|---|---|
| Pre-ablation HEAD | 0.7209 | 0.5336 | 0.8088 |
| v1.0.0-rc | 0.7327 | 0.5476 | 0.8144 |

Ablations C1 (rerank no-surplus skip) and C2 (rerank path-glob searches) shipped after full-suite validation; structural graph augmentation (`VERA_GRAPH_AUGMENT=1`) merged experimental and off by default (+0.0047 nDCG at +83% latency: not worth default-on yet). Detail: `docs/benchmarks.md`.

**Agent-level benchmark** (`benchmarks/agent-bench/`): 10 cross-file Flask questions, agent with Vera vs blocked CLI, blind rubric judging. Kimi K3 used 17% fewer input tokens with Vera (230.6k vs 278.0k) at equal answer quality; Claude Opus 5 scored 10/10 in both arms.

**Bug fixes reported via Discord and review:** lossy UTF-8 file reads, Unicode path glob matching, grep byte-index char-boundary panic, repeatable `--path` filters, ONNX model integrity validation with safe repair, Apple Silicon CoreML memory scaling, embedding model aliases, Voyage reranker endpoints, TS/JS/Java class-method extraction.

**Merges:** 8 community PRs (#22, #25-#31) plus fork fixes; `vera serve` local HTTP inference server; dependency hygiene (RUSTSEC-2026-0204).

### Validation

- `cargo test --workspace`: 910 passed
- `cargo clippy --workspace -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- `cargo audit`, `cargo deny check`, `cargo machete`: clean
- Full Semble suite on the release candidate: see numbers above

### Notes

- Build-from-source now fails loudly with a clear message when the bootstrapped grammars are missing (`scripts/bootstrap-vendored-grammars.sh`).
- Version comes from the git tag at release time; no manifest edits in this PR.

---
## Summary by cubic
Ships v1.0.0 with stronger retrieval, cancellable indexing, git‑scoped and structural search, explain‑path, stale‑index warnings, and a pure inference `vera serve` API. Old behavior: intent text flowed into BM25 and could disable rerank; new behavior: intent only enriches vector queries and always keeps rerank on for intent searches.

- Retrieval: filtered BM25 to prevent scoped‑query starvation; BM25/vector run in parallel; adaptive fusion weights (BM25 2.5x for identifiers); identifier‑aware BM25 with stemming and file‑saturation decay; exact callers/implementations; new structural intents with explicit type‑relation indexing; tightened heuristics and gating; optional graph augmentation via `VERA_GRAPH_AUGMENT=1`.
- Indexing and server: cooperative cancellation and bounded in‑flight remote work; progress reporting and `--max-files`; stale‑index detection and index‑health metrics; lossy UTF‑8 reads; ONNX integrity validation with safe repair; model‑namespaced embedding cache and `model_aliases`; Apple Silicon reranker uses quantized INT8; Potion CPU embedding; `vera serve` honors `return_documents`, cancels on client disconnects, and adds idle‑timeout model caching.
- CLI/MCP: git scopes (`changed`/`since`/`base`), repeatable `--path`, `explain-path`, `structural` intents, exact‑match hints; stricter vendored‑grammar checks; git‑rev validation and atomic updates; MCP indexing progress and cancellation; MCP `search_code` requires `query` and accepts `path` as string or array; agent snippets stay up to date.
- Refactors: remove unused search wrappers and duplicate embed paths; share result/key predicates and presentation helpers; consolidate benchmark metric logic; restructure local model management into validated `assets`/`cuda`/`ort` modules; prune stale artifacts and spikes.

**Rollout and migration**
- Build from source: run `scripts/bootstrap-vendored-grammars.sh` before `cargo build`.
- Reindex projects to populate identifier‑aware BM25 fields, type relations, and index‑health metrics.
- Apple Silicon: run `vera repair --onnx-jina-coreml` to replace any cached fp16 reranker with the quantized model.
- Behavior changes: `--path app/src` matches `app/src/**`; path‑scoped searches always rerank; intent no longer affects BM25 and forces rerank when present.
- MCP integrators: ensure `search_code` sends `query`; `path` may be a string or array.
- Optional: run `vera serve` with `--api-key` or `VERA_SERVE_KEY` to offload embeddings/rerank; enable `VERA_GRAPH_AUGMENT=1` to try graph augmentation.

<sup>Written for commit a8b7dde520d3afbc9562f2c0d6499671babbc24c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/lemon07r/Vera/pull/32?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lemon07r:
@cubic-dev-ai review this

PR #33, opened 2026-08-19
fix(retrieval): run the reranker on CPU under CoreML

### LINKED ISSUE #41: CoreML backend: reranker fails at inference and searches silently return unreranked results
## Summary

On an `onnx-jina-coreml` backend the local reranker fails at inference. Searches do not error: they log a warning and return unreranked results, so the CoreML backend silently loses cross-encoder ranking while still exiting 0.

## Reproduction

Backend `onnx-jina-coreml`, any index.

```
$ vera doctor --probe
  ok    probe-provider-registration registered coreml
  ok    probe-embedding-session embedding session created
  ok    probe-reranker-session reranker session created
  fail  probe-tiny-inference reranker probe failed: Non-zero status code returned while running
        11619352191107991357_CoreML_11619352191107991357_3 node.
        Name:'CoreMLExecutionProvider_11619352191107991357_CoreML_11619352191107991357_3_3'
        Status Message: Error executing model: Unable to compute the prediction using a neural
        network model. It can be an invalid input data or broken/unsupported model (error code: -1).
  skip  probe-provider-confirmation skipped because the tiny inference probe did not succeed
```

Deterministic across repeated runs.

The same failure reaches real searches:

```
$ vera --timing search "cross-encoder reranking of search candidates" --limit 3
WARN vera_core::retrieval::hybrid: reranker unavailable, returning unreranked results
  error=reranker API error (status 500): Non-zero status code returned while running
  ...CoreMLExecutionProvider_..._3_3 ... (error code: -1)
Warning: reranker unavailable (...), returning unreranked results.
[timing] reranking: 166ms
```

Results are returned in fusion order. Exit code 0.

## Root cause

`build_session` in `crates/vera-core/src/retrieval/local_reranker.rs` registers `CoreMLExecutionProvider` whenever the backend is CoreML.

#25 established that no prebuilt reranker export runs on the CoreML GPU and moved every backend onto the quantized INT8 export. The premise recorded with that change is that the CoreML EP can execute none of those ops, so ONNX Runtime partitions zero nodes and falls back to CPU on its own. `vera doctor`'s `probe-reranker-coreml-cpu` warning and the `docs/troubleshooting.md` entry were both written on that basis.

The premise does not hold. The failing node name carries a `CoreMLExecutionProvider_` prefix and a subgraph index, so ORT did assign a fused subgraph to CoreML. Partitioning and CoreML model compilation both succeed, which is why `probe-reranker-session` passes; only the prediction fails.

The existing CPU retry in `LocalReranker::new_with_ep` cannot catch this, because it wraps session construction only and construction succeeds. Inference then runs with no fallback behind it.

Not an ORT version drift: `ORT_VERSION` is `1.24.4` at v0.12.13, at #25's merge commit, and at v1.0.0.

## Impact

Cross-encoder reranking is what lifts MRR@10 from 0.28 to 0.60 in the project's own benchmark table, so this removes most of the ranking advantage over plain hybrid retrieval on Apple Silicon. Because it degrades silently rather than failing, an affected user has no reason to suspect it.

## Environment

Apple M4, macOS 26.6.1, ONNX Runtime 1.24.4, vera 1.0.0 (aarch64-apple-darwin), backend `onnx-jina-coreml`, reranker asset `onnx/model_quantized.onnx`.

## Fix

PR #33 selects the CPU provider for the reranker up front, making the behaviour that `probe-reranker-coreml-cpu` and the troubleshooting docs already describe actually true.

citron07r:
Closes #41.

## Problem

On an `onnx-jina-coreml` backend the reranker fails at inference:

```
Non-zero status code returned while running 11619352191107991357_CoreML_11619352191107991357_3 node.
Name:'CoreMLExecutionProvider_11619352191107991357_CoreML_11619352191107991357_3_3'
Status Message: Error executing model: Unable to compute the prediction using a neural network
model. It can be an invalid input data or broken/unsupported model (error code: -1).
```

`vera doctor --probe` surfaces it as a `probe-tiny-inference` failure, with `probe-reranker-session` still green. Reproduced deterministically on three consecutive runs.

The user-visible effect is worse than the probe suggests. Searches do not fail. They log

```
reranker unavailable, returning unreranked results
```

and return fusion order, so the CoreML backend silently loses cross-encoder ranking while the command still exits 0.

## Root cause

`build_session` in `local_reranker.rs` registered `CoreMLExecutionProvider` whenever the backend was CoreML.

#25 established that no prebuilt reranker export runs on the CoreML GPU and moved every backend onto the quantized INT8 export. The premise recorded alongside that change is that the CoreML EP can execute none of those ops, so ORT partitions zero nodes and falls back to CPU on its own, and `vera doctor`'s `probe-reranker-coreml-cpu` warning was added on that basis.

That premise does not hold here. The failing node name carries a `CoreMLExecutionProvider_` prefix and a subgraph index, so ORT did assign a fused subgraph to CoreML. Partitioning and CoreML model compilation both succeed, which is why `probe-reranker-session` passes; only the prediction fails.

The existing CPU retry cannot catch this. It wraps session construction only, and construction succeeds, so inference runs with no fallback behind it.

This is not an ORT version drift: `ORT_VERSION` is `1.24.4` at v0.12.13, at #25's merge commit, and at v1.0.0.

## Fix

Select the CPU provider for the reranker up front, through a small mapping function placed next to `RERANKER_ONNX_FILE` where the reranker-asset reasoning already lives. This makes the behaviour that `probe-reranker-coreml-cpu` and `docs/troubleshooting.md` already describe actually true.

Embedding still registers CoreML. Only the reranker moves to CPU, which is what the quantized INT8 export is already chosen for.

The mapping is a pure function, so the regression is unit testable without ONNX Runtime or model assets. The pre-existing `test_local_reranker` could not have caught this: it pins `OnnxExecutionProvider::Cpu` and early-returns when ORT is unavailable.

`docs/troubleshooting.md` is corrected to stop asserting that ORT falls back on its own.

## Verification

Apple M4, macOS 26.6.1, ONNX Runtime 1.24.4, backend `onnx-jina-coreml`, index of this repository (375 files, 4790 chunks). Same index and query for both runs.

| | released v1.0.0 | this branch |
|---|---|---|
| `probe-tiny-inference` | fail | ok |
| `probe-provider-confirmation` | skipped | ok |
| search | `reranker unavailable, returning unreranked results` | reranks normally |
| `[timing] reranking` | 166ms (failure path) | 6418ms |

`cargo fmt --check` clean. `cargo clippy -p vera-core --lib` introduces no new warnings; the 5 reported are pre-existing (`pip_package_for_ep`, `CUDA_RUNTIME_LIBRARY_PREFIXES`, `parse_cuda_major_from_runtime_library_entry`, an unused import, an unused variable). New test and the `local_models` suite pass.

## Note

Unrelated to this change and present on both binaries against the same index: `vector search failed, falling back to BM25-only results` with `storage error: vector search failed`. Not addressed here.

---
<a href="https://app.blacksmith.sh/VeraTools/codesmith/Vera/pr/33"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789714830&installation_model_id=432833&pr_number=33&repository=VeraTools%2FVera&return_to=https%3A%2F%2Fgithub.com%2FVeraTools%2FVera%2Fpull%2F33&signature=db0cf35420a3073a7759fdd432e92ff2db221aee18df23c5ffdaae9046043656"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reranker reliability on Apple Silicon by explicitly using CPU execution when CoreML is selected.
  * Added diagnostics for cases where CoreML accepts part of the workload but fails during inference.

* **Documentation**
  * Updated troubleshooting guidance to explain CoreML reranker behavior and the resulting diagnostic warning.


lemon07r:
@coderabbitai review

@cubic-dev-ai review

lemon07r:
Follow-up commit 9b743a2 pushed: the two stale comments (local_models/mod.rs near RERANKER_ONNX_FILE, doctor.rs probe-reranker-coreml-cpu) now describe the explicit CPU pinning via reranker_execution_provider instead of claiming ORT silently falls back on its own.

lemon07r:
@coderabbitai review

PR #42, opened 2026-08-19
fix(storage): clamp KNN k to the sqlite-vec limit instead of losing vector search

### LINKED ISSUE #38: vector search fails for some queries and silently degrades to BM25-only (storage error: vector search failed)
## Summary

On an `onnx-jina-coreml` backend, `vector search failed` for some queries but not others against the same index. Vera degrades to BM25-only results and continues, so the only signal is a WARN line.

```
WARN vera_core::retrieval::hybrid: vector search failed, falling back to BM25-only results
  error=storage error: vector search failed
```

## Reproduction

Index used: a clone of this repository at v1.0.0, 375 files, 4790 chunks, embeddings `jinaai/jina-embeddings-v5-text-nano-retrieval` (fp16), backend `onnx-jina-coreml`.

Counting only the WARN line, one run per query:

| query | vector search failed |
|---|---|
| `fusion` | no |
| `hybrid fusion` | no |
| `hybrid fusion work` | no |
| `how does hybrid fusion work` | **yes** |
| `reranking` | no |
| `cross-encoder reranking of search candidates` | **yes** |

It is deterministic per query across repeated runs, not flaky. Longer, more natural-language queries are the ones that fail, which is the shape of query the tool is designed for.

## Notes

- Reproduces identically on the released v1.0.0 binary and on a local build, so it is not related to any local change.
- The error surfaces at `vector_store.search(...)` in `crates/vera-core/src/retrieval/vector.rs:69`, wrapped as `VectorSearchError::StorageError` with context `"vector search failed"`. The underlying store error is not included in the logged message, so I could not get further without patching.
- `RUST_LOG=debug` did not add detail beyond the same WARN.

## Impact

Silent quality loss. The command exits 0 and returns plausible results, so a user has no reason to suspect the vector half of hybrid retrieval did not run. Combined with reranking being unavailable on this backend (see #33), a CoreML user can be served pure BM25 output while believing they are getting the full pipeline.

## Suggestion

Propagate the underlying store error into the warning so the cause is diagnosable from the log. A hybrid search that loses a whole retrieval arm may also deserve louder treatment than a WARN.

## Environment

Apple M4, macOS 26.6.1, ONNX Runtime 1.24.4, vera 1.0.0, backend `onnx-jina-coreml`.

citron07r:
Closes #38.

## Root cause

`VectorStore::search` passes its `limit` straight into the SQL `LIMIT`, which sqlite-vec reads as the KNN `k` and hard-rejects above 4096:

```
k value in knn query too large, provided 4260 and the limit is 4096: Error code 1: SQL logic error
```

Callers scale the candidate pool from the query type and result limit, in three compounding steps:

1. `compute_vector_candidates(limit, multiplier)` in `retrieval/hybrid.rs`
2. `run_hybrid_search` multiplies again by 4 for natural-language queries (`hybrid.rs:135`)
3. `vector_search_with_stores` adds a further 50% overfetch (`retrieval/vector.rs:64`)

On a 4790-chunk index a five-word query reached 4260. The vector arm was then dropped entirely and only BM25 results were returned, for exactly the conversational queries hybrid retrieval exists to serve.

## Why it was invisible

`VectorSearchError::StorageError` formats its inner `anyhow::Error` with `{0}`, which prints only the outermost context and discards the cause chain. Every failure surfaced as the useless:

```
error=storage error: vector search failed
```

Switching to `{0:#}` exposes the real cause. That one-character change is what made this diagnosable at all, so it is included here.

## Change

- Clamp `k` to `MAX_KNN_K` (4096) in `VectorStore::search`. Requesting 4096 candidates rather than 4260 is imperceptible; losing vector search entirely is not.
- Format the storage error with `{0:#}` so future failures are diagnosable from the log rather than requiring a patched build.

I put the clamp at the storage layer because 4096 is a property of the sqlite-vec backend, not of any caller's pool arithmetic. Clamping there fixes every current and future caller at once.

## Verification

Repeated runs on a 4790-chunk index, counting only the fallback warning:

| query | v1.0.0 | this branch |
|---|---|---|
| `fusion` | ok | ok |
| `hybrid fusion work` | ok | ok |
| `how does hybrid fusion work` | **BM25-only** | ok |
| `cross-encoder reranking of search candidates` | **BM25-only** | ok |

`search_clamps_limit_above_sqlite_vec_knn_cap` fails without the clamp with the production error (`k value in knn query too large, provided 4097 and the limit is 4096`) and passes with it. All 13 `storage::vector` tests pass. `cargo fmt --check` clean.

## Note

Clamping keeps a large pool silently smaller than requested. If you would rather callers not build oversized pools in the first place, the alternative is to bound the multipliers in `hybrid.rs` against index size and keep the storage layer strict. Happy to switch. The error-format change is worth having either way.

---
<a href="https://app.blacksmith.sh/VeraTools/codesmith/Vera/pr/42"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789717390&installation_model_id=432833&pr_number=42&repository=VeraTools%2FVera&return_to=https%3A%2F%2Fgithub.com%2FVeraTools%2FVera%2Fpull%2F42&signature=3978d6d4d206977c06b5ee31e7a2f6d50a17788a6ade5f2a621a09f4ec2e210a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/vera-core/src/retrieval/vector.rs
+++ b/crates/vera-core/src/retrieval/vector.rs
@@ -25,7 +25,7 @@ pub enum VectorSearchError {
     },
 
     /// Storage or metadata error.
-    #[error("storage error: {0}")]
+    #[error("storage error: {0:#}")]
     StorageError(#[from] anyhow::Error),
 }
 
--- a/crates/vera-core/src/retrieval/vector.rs
+++ b/crates/vera-core/src/retrieval/vector.rs
@@ -171,6 +171,7 @@ mod tests {
     use super::*;
     use crate::embedding::test_helpers::MockProvider;
     use crate::types::{Chunk, Language, SymbolType};
+    use anyhow::Context;
 
     /// Create sample chunks with semantic variety for testing.
     fn sample_chunks() -> Vec<Chunk> {
@@ -316,6 +317,15 @@ mod tests {
                 scores[i],
             );
         }
+    }
+
+    #[test]
+    fn storage_error_display_includes_error_chain() {
+        let error = anyhow::anyhow!("root cause").context("outer context");
+        let display = VectorSearchError::StorageError(error).to_string();
+
+        assert!(display.contains("outer context"));
+        assert!(display.contains("root cause"));
     }
 
     // ── generate_query_embedding tests ───────────────────────────────
--- a/crates/vera-core/src/storage/vector.rs
+++ b/crates/vera-core/src/storage/vector.rs
@@ -207,6 +207,13 @@ impl VectorStore {
         // it on natural-language queries, and the whole vector arm would then be
         // dropped in favour of BM25-only results. Ask for as many as the backend
         // allows instead.
+        if limit > MAX_KNN_K {
+            tracing::debug!(
+                requested = limit,
+                clamped = MAX_KNN_K,
+                "clamping vector search limit to sqlite-vec KNN cap"
+            );
+        }
         let limit = limit.min(MAX_KNN_K);
 
         let mut stmt = self
--- a/crates/vera-core/src/storage/vector.rs
+++ b/crates/vera-core/src/storage/vector.rs
@@ -14,6 +14,10 @@ pub struct VectorStore {
     dim: usize,
 }
 
+/// Maximum `k` sqlite-vec accepts in a KNN query. Requesting more is a hard
+/// error from the extension, not a soft limit.
+const MAX_KNN_K: usize = 4096;
+
 /// A single vector search result: chunk ID and distance score.
 #[derive(Debug, Clone)]
 pub struct VectorSearchResult {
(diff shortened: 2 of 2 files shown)


lemon07r:
@coderabbitai review

@cubic-dev-ai review

lemon07r:
Follow-up commit 0a52bbd pushed, addressing the remaining review nits (one also raised by cubic):

1. The KNN clamp now emits a tracing::debug log with requested vs clamped values, so large-request truncation is diagnosable rather than silent.
2. New test storage_error_display_includes_error_chain pins the {0:#} cause-chain formatting.

45 vera-core vector tests pass; clippy/fmt clean.

lemon07r:
@cubic-dev-ai review

lemon07r:
@coderabbitai review

lemon07r:
Noted on the P3. The clamp's diagnosability gap is covered by the tracing::debug log added in 0a52bbd (requested vs clamped k). Bounding the multipliers upstream in hybrid.rs against index size is a reasonable future cleanup, but the storage clamp stays as the backend-contract backstop either way, so leaving as-is for this PR.

Issue #35, opened 2026-08-19
vera upgrade --apply cannot install v1.0.0: npm and PyPI wrappers still publish 0.12.13


citron07r:
## Summary

v1.0.0 exists on GitHub Releases but was never published to npm or PyPI, so `vera upgrade --apply` cannot reach it and silently reinstalls the version already present.

## Reproduction

```
$ vera upgrade
Current version: 0.12.13
Latest version:  1.0.0
Update status:   update available
Install method:  bun (provenance)
Update command:  bun install -g @vera-ai/cli@latest && bunx @vera-ai/cli install
Apply support:   yes (`vera upgrade --apply`)

$ vera upgrade --apply
installed @vera-ai/cli@0.12.13 with binaries:
 - cli
Vera 0.12.13 installed.
```

Registry state:

```
$ curl -s https://registry.npmjs.org/@vera-ai/cli | jq .'dist-tags'.latest
"0.12.13"
$ curl -s https://pypi.org/pypi/vera-ai/json | jq -r .info.version
0.12.13
$ curl -s https://api.github.com/repos/VeraTools/Vera/releases/latest | jq -r .tag_name
v1.0.0
```

## Impact

`vera doctor` advertises an update that no documented install path can deliver. Every wrapper is affected, since `bunx`/`npx`/`uvx` all resolve through the same two registries. The upgrade reports success while leaving the old binary in place, which is the worst of the failure modes: no error to act on.

The workaround is a manual install of `vera-aarch64-apple-darwin.tar.gz` from the release, verifying against `release-manifest.json` and repointing the shim plus `~/.vera/install.json` by hand.

## Suggestions

Either publish the wrappers as part of the release, or have `vera upgrade` compare the version actually available from the resolved install method and report when the registry has not caught up, rather than reporting the GitHub tag as reachable.

## Environment

Apple M4, macOS 26.6.1. Checked 2026-08-19, two days after the v1.0.0 tag.

citron07r:
The cause is a failed release job, not a forgotten manual step.

The **Release** workflow run for `v1.0.0` ([run 32069441751](https://github.com/VeraTools/Vera/actions/runs/32069441751)) is red. All six `Build` matrix jobs succeeded; `Create Release` failed:

```
npm error code E404
npm error 404 Not Found - PUT https://registry.npmjs.org/@vera-ai%2fcli - Not found
npm error 404
npm error 404  '@vera-ai/cli@1.0.0' is not in this registry.
[...]

PR #43, opened 2026-08-19
fix(upgrade): verify the applied version instead of trusting the installer exit code

### LINKED ISSUE #44: vera upgrade --apply reports success when the installer ran but the version did not change
## Summary

`vera upgrade --apply` reports success when the installer ran but the version did not change. It checks the installer command's exit code and nothing else.

## Reproduction

Any situation where `<pkg>@latest` resolves to the version already installed, for example a package registry lagging behind a GitHub release:

```
$ vera upgrade
Current version: 0.12.13
Latest version:  1.0.0
Update status:   update available
Install method:  bun (provenance)
Update command:  bun install -g @vera-ai/cli@latest && bunx @vera-ai/cli install
Apply support:   yes (`vera upgrade --apply`)

$ vera upgrade --apply
installed @vera-ai/cli@0.12.13 with binaries:
 - cli
Vera 0.12.13 installed.
```

Exit code 0. `applied` is reported as true. The binary is unchanged, and the next `vera doctor` advertises the same update again.

## Cause

In `crates/vera-cli/src/commands/upgrade.rs`:

```rust
update_check::apply_update(method)?;
report.applied = true;
print_report(&report, json_output)
```

`apply_update` shells out to the package manager and propagates its exit status. Every command in the chain exits 0 when `@latest` already matches the installed version, so nothing distinguishes a real upgrade from a no-op.

## Impact

This is a silent-success failure: no error to act on, and the mismatch between "update available" and "nothing changed" is only visible if the user manually re-checks the version afterwards. It masks whatever upstream problem prevented the new version from being installable, which is how #35 stayed invisible.

## Suggestion

Compare the version the installer recorded in `install.json` against the version that was advertised, and fail with both numbers when they differ. `state::load_install_provenance()` already exposes it.

PR #43 implements this.

citron07r:
Closes #44. Addresses the client-side half of #35, which stays open for the publishing failure itself.

## Problem

`vera upgrade --apply` reports success while leaving the old binary in place:

```
$ vera upgrade
Current version: 0.12.13
Latest version:  1.0.0
Update status:   update available
Apply support:   yes (`vera upgrade --apply`)

$ vera upgrade --apply
installed @vera-ai/cli@0.12.13 with binaries:
 - cli
Vera 0.12.13 installed.
```

The v1.0.0 release workflow failed at the npm publish step ([run 32069441751](https://github.com/VeraTools/Vera/actions/runs/32069441751)), so the registry still serves 0.12.13 and `@vera-ai/cli@latest` resolves to the version already installed. Every installer command exits 0, `applied` is set, and the report claims success.

This is the worst failure mode available: no error to act on, and `vera doctor` keeps advertising an update that upgrading appears to complete.

## Change

`run` now compares the version the installer recorded in `install.json` against the version that was advertised:

- **Mismatch** → fail, naming both versions and the likely cause (the package for that version has not been published, so `@latest` still resolves to the old one), and pointing at the GitHub release as an alternative.
- **Installed version unreadable** → say so rather than implying the upgrade was confirmed.
- **Match** → unchanged.

`UpgradeReport` gains an `installed_version` field so `--json` consumers see the same evidence.

The comparison is a pure helper, `applied_version_mismatch`, so it is unit tested without running an installer.

## Scope

This does not fix the publishing failure itself, which needs `NPM_TOKEN` attention on the maintainer side (details in #35). It makes the client stop reporting a no-op as a success.

## Verification

`vera upgrade` dry-run output is unchanged. Three new tests cover match, mismatch, and unknown-version cases; all 85 `vera-cli` tests pass. `cargo fmt --check` clean, and `cargo clippy -p vera-cli --bin vera` adds no new warnings (the 5 reported originate in `vera-core` and predate this change).

## Possible follow-up

Verifying against the running binary's own version after re-exec, rather than the recorded `install.json` value, would also catch an installer that writes provenance without replacing the binary. That seemed like more machinery than this warrants; happy to add it if you would rather.

---
<a href="https://app.blacksmith.sh/VeraTools/codesmith/Vera/pr/43"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789717851&installation_model_id=432833&pr_number=43&repository=VeraTools%2FVera&return_to=https%3A%2F%2Fgithub.com%2FVeraTools%2FVera%2Fpull%2F43&signature=c4fef4cf9e4250923f28b317c43074952c083dd6c75ca2ef92648fac1f6765bd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/vera-cli/src/commands/upgrade.rs
+++ b/crates/vera-cli/src/commands/upgrade.rs
@@ -57,7 +57,6 @@ pub fn run(apply: bool, json_output: bool) -> Result<()> {
         .as_deref()
         .expect("apply requires a resolved install method");
     update_check::apply_update(method)?;
-    report.applied = true;
 
     // The installer command exiting 0 does not mean the new version landed. A
     // package registry can lag behind a GitHub release, in which case
@@ -68,17 +67,24 @@ pub fn run(apply: bool, json_output: bool) -> Result<()> {
         .ok()
         .and_then(|provenance| provenance.version);
 
-    if report.installed_version.is_none() && !json_output {
-        eprintln!("Note: could not read the installed version to confirm the upgrade applied.");
-    }
-
-    if let Some(message) = applied_version_mismatch(
+    match verification_outcome(
         method,
         report.latest_version.as_deref(),
         report.installed_version.as_deref(),
     ) {
-        print_report(&report, json_output)?;
-        bail!(message);
+        VerificationOutcome::Confirmed => report.applied = true,
+        VerificationOutcome::Mismatch(message) => {
+            print_report(&report, json_output)?;
+            bail!(message);
+        }
+        VerificationOutcome::Unknown => {
+            let message = "could not confirm the upgrade applied because the installed version is unavailable";
+            if !json_output {
+                eprintln!("Warning: {message}.");
+            }
+            print_report(&report, json_output)?;
+            bail!(message);
+        }
     }
 
     print_report(&report, json_output)
@@ -108,9 +114,31 @@ fn applied_version_mismatch(
     ))
 }
 
+#[derive(Debug, PartialEq, Eq)]
+enum VerificationOutcome {
+    Confirmed,
+    Mismatch(String),
+    Unknown,
+}
+
+fn verification_outcome(
+    method: &str,
+    latest: Option<&str>,
+    installed: Option<&str>,
+) -> VerificationOutcome {
+    match (latest, installed) {
+        (Some(latest), Some(installed)) if latest == installed => VerificationOutcome::Confirmed,
+        (Some(latest), Some(installed)) => VerificationOutcome::Mismatch(
+            applied_version_mismatch(method, Some(latest), Some(installed))
+                .expect("known, mismatched versions must produce a mismatch message"),
+        ),
+        _ => VerificationOutcome::Unknown,
+    }
+}
+
 #[cfg(test)]
 mod tests {
-    use super::applied_version_mismatch;
+    use super::{VerificationOutcome, applied_version_mismatch, verification_outcome};
 
     #[test]
     fn reports_nothing_when_the_installed_version_matches() {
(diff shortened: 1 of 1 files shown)


lemon07r:
@coderabbitai review

@cubic-dev-ai review

lemon07r:
Follow-up commit 72229aa pushed, addressing the two review findings (also flagged by cubic):

1. **Mismatch path reported `applied: true`**: `report.applied` is now set only after the recorded installed version matches the advertised version. Mismatch prints the report with `applied: false` and exits nonzero.
2. **Unreadable/missing installed version implied success**: now maps to an explicit unconfirmed state — warning on text output, `"applied": false` in JSON, nonzero exit so scripts can catch it.

Both behaviors are covered by new unit tests on a pure `verification_outcome` helper; the existing `applied_version_mismatch` tests are unchanged.

lemon07r:
@cubic-dev-ai review

lemon07r:
@coderabbitai review

lemon07r:
Both follow-ups addressed in 7e72ef5: verification_outcome now delegates to applied_version_mismatch (single equality check, no expect), and the mismatch message is package-manager-neutral (no @latest assumption; also points at the VeraTools org releases page).

@cubic-dev-ai review

PR #49, opened 2026-08-19
ci: speed up release builds, fix rocm docker publish

lemon07r:
## Why

v1.0.1 on Blacksmith exposed two problems:

1. **Speed/balance**: release wall-clock was 14.3 min, bounded by the slowest 2vcpu job while faster jobs idled. Docker took 25+ min per variant because every image recompiled the entire workspace in-container.
2. **Correctness**: the rocm publish failed twice with `failed commit on ref "layer-sha256:..." unexpected commit digest` — the same layer ref both times, i.e. a poisoned entry in the persistent builder's cache, not a flake. `1.0.1-rocm` is currently missing from GHCR.

## Runner rebalancing (measured v1.0.1 durations)

| Target | Was (2vcpu unless noted) | Now | Est. |
|--------|--------------------------|-----|------|
| aarch64-linux-gnu | 856s (critical path) | 8vcpu-arm | ~295s |
| windows-msvc | 749s | 8vcpu | ~258s |
| musl | 633s | 4vcpu | ~372s (apt setup floors it; new critical path) |
| linux-gnu | 554s | 4vcpu | ~326s |
| macOS aarch64 | 166s @6vcpu | 6vcpu (unchanged) | ~166s |
| macOS x86_64 | 156s @6vcpu | 6vcpu (unchanged) | ~156s |

Everything lands in a ~2.7-6.2 min band instead of 2.6-14.3 min. No smaller macOS label exists (`blacksmith-4vcpu-macos-15` smoke-tested: never picked up), and shrinking macOS below 6vcpu wouldn't help wall-clock anyway since it was never the bottleneck. Docker's compile job runs on 8vcpu (~3 min), so its total (~5 min) also stays under the musl path.

## Docker changes

- New `binary` job compiles `vera` once (8vcpu, sccache + rust-cache); the three package jobs download it and only pull the base image + `COPY dist/vera`. The variants already built byte-identical binaries (same `cargo build --release`, no feature flags), so image behavior is unchanged.
- `cache-key: docker-v2-<variant>` provisions fresh builders, leaving the poisoned rocm cache behind.
- `workflow_dispatch` with a `tag` input to republish images for an existing tag without re-tagging; binary builds from the tag, packaging uses the workflow's own ref. Used to backfill `1.0.1-rocm` after merge.
- Concurrency group serializes publishes so a manual republish can't race a tag push on the mutable `:cpu/:cuda/:rocm` tags.
- Tag value passed via `env:` instead of inline template expansion (script-injection hardening).

## Dockerfiles (cpu, cuda, rocm, openvino)

Builder stage removed; each packages `dist/vera`. Headers and `docs/docker.md` document the local flow: `bootstrap-vendored-grammars.sh` (needed on fresh clones), `cargo build --release -p vera-cli`, copy to `dist/`, then `docker build`.

## Validation

- actionlint clean; YAML parses.
- Smoke runs on every new label (`4vcpu-ubuntu-2404`, `4vcpu-ubuntu-2404-arm`, `8vcpu-ubuntu-2404`, `8vcpu-ubuntu-2404-arm`, `8vcpu-windows-2025`): all picked up and succeeded. `4vcpu-macos-15` does not exist.
- Full publish path exercised right after merge via the new dispatch trigger (`1.0.1-rocm` backfill).

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -3,5 +3,9 @@ self-hosted-runner:
   labels:
     - blacksmith-2vcpu-ubuntu-2404
     - blacksmith-2vcpu-ubuntu-2404-arm
+    - blacksmith-4vcpu-ubuntu-2404
+    - blacksmith-4vcpu-ubuntu-2404-arm
+    - blacksmith-8vcpu-ubuntu-2404
     - blacksmith-6vcpu-macos-15
     - blacksmith-2vcpu-windows-2025
+    - blacksmith-4vcpu-windows-2025
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -6,6 +6,8 @@ self-hosted-runner:
     - blacksmith-4vcpu-ubuntu-2404
     - blacksmith-4vcpu-ubuntu-2404-arm
     - blacksmith-8vcpu-ubuntu-2404
+    - blacksmith-8vcpu-ubuntu-2404-arm
     - blacksmith-6vcpu-macos-15
     - blacksmith-2vcpu-windows-2025
     - blacksmith-4vcpu-windows-2025
+    - blacksmith-8vcpu-windows-2025
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,5 +1,11 @@
 name: Docker
 
+# Serialize image publication: a manual republish and a tag push must not
+# race on the same mutable :cpu/:cuda/:rocm tags.
+concurrency:
+  group: docker-image-publication
+  cancel-in-progress: false
+
 on:
   push:
     tags: ['v*.*.*']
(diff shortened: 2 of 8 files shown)


lemon07r:
@coderabbitai review

lemon07r:
@cubic-dev-ai review

lemon07r:
CodeRabbit review disposition:

1. **Concurrency group** (docker.yml:19): already fixed in 93eaa79 (workflow-level group, cancel-in-progress: false).
2. **Script injection via template expansion** (docker.yml:36, :54): already fixed in 93eaa79 (tag passed via env:).
3. **ARM runner label** (release.yml:30): false positive. The aarch64 target requires an ARM runner; x86_64 cannot produce it natively. Current code uses blacksmith-8vcpu-ubuntu-2404-arm, smoke-verified to schedule and succeed.
4. **Prebuilt binary platform** (docs/docker.md): valid, fixed in 993dbc5 (Linux x86_64 requirement noted in docs and Dockerfile headers).

PR #58, opened 2026-08-19
fix(parsing): parse .tsx with the tsx grammar and index JSX call sites

### LINKED ISSUE #56: bug: .tsx parsed with the typescript grammar, so JSX call sites are invisible to `references` and `dead-code`
`.tsx` files appear to be parsed with the `typescript` tree-sitter grammar rather than `tsx`. Because JSX is ambiguous with type assertions, the two grammars are separate upstream, and using the TypeScript one turns every JSX element into an error node.

The visible symptom is an error count. The real cost is that `vera references` returns a confident **"No callers found"** for React components that do have callers, because the call sites are JSX elements sitting inside those error nodes.

## Environment

- vera 1.0.0 (`aarch64-apple-darwin`, local build)
- macOS, Apple Silicon, CoreML local mode
- Reproduced on a React 19 + TypeScript + Vite codebase

## Reproduction

Four files, each indexed in its own directory:

```ts
// plain.ts
export function add(a: number, b: number): number { return a + b; }
export interface User { id: string; name: string; }
```
```ts
// modern.ts
const conf = { a: 1 } satisfies Record<string, number>;
export const f = <const T,>(x: T) => x;
using res = { [Symbol.dispose]() {} };
```
```tsx
// nojsx.tsx
export const VALUE = 42;
export type Thing = { a: string };
```
```tsx
// jsx.tsx
export function Hello({ name }: { name: string }) {
  return <div className="greet">Hello {name}</div>;
}
```

`vera stats --json | .index_health.files_with_tree_sitter_errors`:

| File | Errors |
|------|--------|
| `plain.ts` | 0 |
| `modern.ts` | 0 |
| `nojsx.tsx` | 0 |
| **`jsx.tsx`** | **1** |

JSX is the only trigger. Notably `modern.ts` is clean, so this is not a general "grammar is behind modern TypeScript" problem: `satisfies`, explicit resource management (`using` / `Symbol.dispose`) and const type parameters all parse fine.

## Impact at repository scale

On a mid-sized React codebase, roughly half of all files reported as language `typescript` carry tree-sitter errors, with **0 parse failures**, so nothing is dropped from the index.

Both `.ts` and `.tsx` are reported under the single language label `typescript`, which is what suggests one grammar is used for both extensions. The errored-file count tracks the number of `.tsx` files that actually contain JSX: `.tsx` files with no JSX in them parse clean.

## Why this matters more than the error count suggests

Minimal end-to-end case, using the `jsx.tsx` above plus one consumer:

```tsx
// app.tsx
import { Hello } from "./jsx";
export function App() {
  return <Hello name="world" />;
}
```

```
$ vera references Hello
No callers found for 'Hello'.
```

The caller exists at `app.tsx`. It is a JSX element, exactly the syntax inside the error node.

**Control, confirming the scope is JSX-specific rather than general breakage:** a plain function called normally, for example `add(1, 2)` from a `.tsx` file, resolves correctly. So `references` works in `.tsx`; it is JSX element usage specifically that is invisible.

The failure mode is the concerning part. `references` and `dead-code` do not report reduced confidence, they report a clean negative. Anyone using them to decide whether a component is unused would be told, incorrectly, that it is.

## Suggested fix

Select the tree-sitter grammar by extension, `tsx` for `.tsx` and `typescript` for `.ts`, rather than mapping both to one language. Upstream tree-sitter ships them as separate grammars for this reason. Both currently appear under one `typescript` label in `vera stats`, so the language breakdown would also become more accurate.

Happy to test a patch against a JSX-heavy codebase.

## Workaround for anyone hitting this

Use `vera grep` or `git grep` to find component usages, and do not trust `vera references` or `vera dead-code` for React components. `vera search` is unaffected, since semantic search does not depend on the syntax tree.

## Note on the fix shape

Selecting the grammar by extension is necessary but not sufficient. `is_call_node`
in `crates/vera-core/src/parsing/references.rs` recognises only `call_expression`
and `new_expression` for TS/JS, so a JSX element is not treated as a call site even
once it parses cleanly. Both halves are needed before `vera references Hello`
returns the caller.

The same gap affects `.jsx`/`.js`, where the JavaScript grammar already parses JSX
without errors and JSX call sites are still invisible.

---

Re-filed from #53, which was opened from the wrong account.

citron07r:
Fixes #56.

`.tsx` was parsed with the `typescript` tree-sitter grammar. JSX is ambiguous with
type assertions, which is why tree-sitter ships `typescript` and `tsx` as separate
grammars upstream, so every JSX element landed in an error node and the call sites
inside it were never parsed.

The visible symptom was a tree-sitter error count. The cost was that `references`
and `dead-code` returned a confident `No callers found` for components rendered as
`<Foo />` — not reduced confidence, a clean negative, which is the wrong answer to
act on when deciding whether a component is dead.

## What changed

**1. Grammar selection by extension.** `tree_sitter_grammar_for_path` returns the
`tsx` grammar for `.tsx` and defers to the existing per-language mapping for
everything else. It sits beside `tree_sitter_grammar` in `parsing/languages.rs`
and is used from `parse_file_with_diagnostics` and the `structural` syntax filter,
both of which already have the file path.

**2. JSX elements as call sites.** Grammar selection alone does not close the issue.
`is_call_node` recognised only `call_expression` and `new_expression` for TS/JS, so
a cleanly parsed JSX element still was not a call site. `jsx_opening_element` and
`jsx_self_closing_element` are now call nodes; `jsx_closing_element` deliberately is
not, so `<Foo></Foo>` records one reference rather than two.

This half also fixes `.jsx`/`.js`, where the JavaScript grammar already parsed JSX
without errors and the call sites were invisible anyway.

**Host elements stay out of the reference graph.** JSX resolves a tag beginning with
a lowercase letter to an intrinsic element string rather than to a value in scope,
so recording `<div>` would add an edge per HTML tag in the repository.

The `typescript` label in `stats` is unchanged. Splitting it would change reported
language breakdowns, which reads as a separate maintainer call.

## Verification

Against the released v1.0.0 binary as a control, same fixture and same queries. Two
files: `jsx.tsx` defining `Hello`, `app.tsx` rendering `<Hello name="world" />`.

| | stock v1.0.0 | this branch |
|---|---|---|
| `vera references Hello` | `No callers found for 'Hello'.` | `app.tsx:1-4 function:App` |
| `files_with_tree_sitter_errors` | 2 | 0 |
| `files_with_parse_failures` | 0 | 0 |
| `vera references div` | `No callers found` | `No callers found` |

Tests: `cargo test --workspace` — 920 passed, 0 failed. `cargo build --release`
clean for the whole workspace. `cargo fmt --check` clean. `cargo clippy -p vera-core
--lib` reports only the five warnings already present on `master`.

Both halves were confirmed load-bearing by reverting each in turn:

- reverting the `is_call_node` change fails `tsx_jsx_elements_are_call_sites` and
  `jsx_in_javascript_is_a_call_site` with the production symptom, an empty callee
  list
- reverting the grammar selection fails `tsx_files_parse_jsx_without_errors` and
  also `tsx_jsx_elements_are_call_sites`, since the JSX nodes do not exist to match

`tsx_files_parse_jsx_without_errors` also asserts that the `typescript` grammar
still errors on the same JSX source, so the test starts failing if the upstream
split ever stops being necessary.

## Deliberately out of scope

`signatures::extract_signature(content, lang)` takes no path, so it still resolves
the grammar by language alone. Its callers do have a path available — `Chunk` and
`SearchResult` both carry `file_path` — so a `.tsx` chunk can still hit JSX in an
error node when a signature is stripped for `--compact` output or a type-relation
header.

That is the same class of problem, but fixing it means changing a public signature
used from `vera-mcp` and `vera-cli` as well as `vera-core`, which is wider than this
issue. Happy to do it here or in a follow-up, whichever you prefer.

Happy to adjust the shape if you would rather see the language label split too, or
the JSX handling scoped differently.

---
<a href="https://app.blacksmith.sh/VeraTools/codesmith/Vera/pr/58"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789743571&installation_model_id=432833&pr_number=58&repository=VeraTools%2FVera&return_to=https%3A%2F%2Fgithub.com%2FVeraTools%2FVera%2Fpull%2F58&signature=3d7bd0da2461260aa48cc7f8e2e844162d4f6eaeadb7a09b9f7143bb23fdacd2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Parse `.tsx` with the `tsx` grammar and record JSX component renders as call sites so `references`/`dead-code` find component usage. Previously `.tsx` used the `typescript` grammar (JSX parsed as errors) and JSX renders like `<Foo />` were invisible.

- Grammar by path: `tree_sitter_grammar_for_path` returns `tsx` for `.tsx` and is used in `parse_file_with_diagnostics` and structural search’s `SyntaxFilter`.
- JSX as calls: treat `jsx_opening_element` and `jsx_self_closing_element` as calls; exclude `jsx_closing_element`. Also surfaces JSX calls in `.js`/`.jsx`.
- Host-element filter: decide from the tag node; skip bare lowercase ASCII tags or any tag containing a hyphen (e.g., `<div>`, `<My-element />`). Keep dotted tags (e.g., `<Icons.arrow />`) and non-ASCII-leading names (e.g., `<éWidget />`) as component calls.
- Structural search: selects grammar by file path; behavior and attribution on `.tsx` are unchanged. Added a characterization test that finds real env reads and filters comment/JSX-attribute decoys.
- Tests cover grammar selection, JSX call extraction (dotted, ASCII-only lowercase, hyphen rules), `.tsx` structural search. No API or `stats` label changes.

<sup>Written for commit 8c73e5a0bc74a4a054efb488539089397963e7c9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/58?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved JavaScript and TypeScript JSX support, including `.tsx` files.
  * JSX components are now recognized in extracted references and call sites.
  * JSX host elements are excluded from component reference results.
  * Structural searches now work reliably in JSX-containing TypeScript files.

* **Bug Fixes**
  * Fixed parsing and structural matching for TSX files by selecting the appropriate grammar based on file path.
  * Added coverage for JSX components, dotted tags, caller detection, environment-variable reads, and clean TSX parsing.

### DIFF
--- a/crates/vera-core/src/parsing/languages.rs
+++ b/crates/vera-core/src/parsing/languages.rs
@@ -106,6 +106,26 @@ pub fn tree_sitter_grammar(lang: Language) -> Option<TsLanguage> {
     Some(lang_fn)
 }
 
+/// Returns the tree-sitter grammar to use for a specific file.
+///
+/// TypeScript ships as two grammars upstream because JSX is ambiguous with
+/// type assertions, so `.tsx` must be parsed with `tsx`. Parsing it with
+/// `typescript` puts every JSX element in an error node, which hides JSX call
+/// sites from `references` and `dead-code`. All other languages resolve by
+/// language alone.
+pub fn tree_sitter_grammar_for_path(lang: Language, file_path: &str) -> Option<TsLanguage> {
+    if lang == Language::TypeScript && has_tsx_extension(file_path) {
+        return Some(tree_sitter_typescript::LANGUAGE_TSX.into());
+    }
+    tree_sitter_grammar(lang)
+}
+
+fn has_tsx_extension(file_path: &str) -> bool {
+    std::path::Path::new(file_path)
+        .extension()
+        .is_some_and(|ext| ext.eq_ignore_ascii_case("tsx"))
+}
+
 /// Returns whether a language has tree-sitter grammar support (Tier 1A).
 pub fn has_grammar(lang: Language) -> bool {
     tree_sitter_grammar(lang).is_some()
@@ -183,6 +203,48 @@ mod tests {
             assert!(
                 !has_grammar(lang),
                 "{lang} should NOT have a tree-sitter grammar"
+            );
+        }
+    }
+
+    /// A `.tsx` file must parse without error nodes; the `typescript` grammar
+    /// puts every JSX element in one, which is what hides JSX call sites.
+    #[test]
+    fn tsx_files_parse_jsx_without_errors() {
+        let source = "export function Hello() {\n  return <div className=\"greet\">hi</div>;\n}\n";
+
+        let tsx = tree_sitter_grammar_for_path(Language::TypeScript, "src/jsx.tsx").unwrap();
+        let mut parser = tree_sitter::Parser::new();
+        parser.set_language(&tsx).unwrap();
+        assert!(
+            !parser.parse(source, None).unwrap().root_node().has_error(),
+            "tsx grammar should parse JSX cleanly"
+        );
+
+        let ts = tree_sitter_grammar_for_path(Language::TypeScript, "src/plain.ts").unwrap();
+        let mut parser = tree_sitter::Parser::new();
+        parser.set_language(&ts).unwrap();
+        assert!(
+            parser.parse(source, None).unwrap().root_node().has_error(),
+            "typescript grammar is the one that errors on JSX; if this stops \
+             holding, the tsx split is no longer needed"
+        );
+    }
+
+    #[test]
+    fn non_tsx_paths_keep_their_language_grammar() {
+        let cases = [
+            ("src/plain.ts", Language::TypeScript),
+            ("src/app.mts", Language::TypeScript),
+            ("noext", Language::TypeScript),
+            ("src/main.rs", Language::Rust),
+        ];
+        for (path, lang) in cases {
+            let by_path = tree_sitter_grammar_for_path(lang, path).unwrap();
+            assert_eq!(
+                by_path,
+                tree_sitter_grammar(lang).unwrap(),
+                "{path} should use the plain {lang} grammar"
             );
         }
     }
--- a/crates/vera-core/src/parsing/mod.rs
+++ b/crates/vera-core/src/parsing/mod.rs
@@ -100,7 +100,7 @@ pub fn parse_file_with_diagnostics(
         ));
     }
 
-    let grammar = match languages::tree_sitter_grammar(language) {
+    let grammar = match languages::tree_sitter_grammar_for_path(language, file_path) {
         Some(g) => g,
         None => {
             let chunks = chunker::tier0_line_chunks(source, file_path, language);
--- a/crates/vera-core/src/parsing/references.rs
+++ b/crates/vera-core/src/parsing/references.rs
@@ -95,12 +95,28 @@ fn is_jsx_element_node(kind: &str) -> bool {
 
 /// Whether a JSX tag names a host element rather than a component in scope.
 ///
-/// JSX resolves a tag beginning with a lowercase letter, such as `<div>`, to an
-/// intrinsic element string, and any other, such as `<Foo>` or `<Foo.Bar>`, to a
-/// value in scope. Only the latter is a call site; recording the former would
-/// add an edge per HTML tag in the repository.
-fn is_jsx_host_element(kind: &str, name: &str) -> bool {
-    is_jsx_element_node(kind) && name.starts_with(|ch: char| ch.is_lowercase())
+/// JSX resolves a *bare* tag beginning with a lowercase letter, such as `<div>`,
+/// to an intrinsic element string. Anything else is a value in scope, and only
+/// those are call sites; recording host elements would add an edge per HTML tag
+/// in the repository.
+///
+/// The decision has to come from the tag node, not from the extracted callee.
+/// `extract_callee` returns the rightmost identifier, so `<Icons.arrow />` yields
+/// `arrow`, and judging that name alone would discard a real component
+/// reference. A dotted tag is a member expression on a value in scope whatever
+/// the case of its property.
+fn is_jsx_host_element(node: &tree_sitter::Node, source: &[u8]) -> bool {
+    if !is_jsx_element_node(node.kind()) {
+        return false;
+    }
+    let Some(name) = node.child_by_field_name("name") else {
+        return false;
+    };
+    if name.kind() == "member_expression" {
+        return false;
+    }
+    name.utf8_text(source)
+        .is_ok_and(|text| text.starts_with(|ch: char| ch.is_lowercase()))
 }
 
 /// Extract the callee name from a call node.
@@ -209,7 +225,7 @@ fn collect_calls(
         let node = cursor.node();
         if is_call_node(lang, node.kind()) {
             if let Some(callee) = extract_callee(&node, source) {
-                if !is_jsx_host_element(node.kind(), &callee) {
+                if !is_jsx_host_element(&node, source) {
                     let caller = find_enclosing_symbol(symbols, node.start_byte());
                     refs.push(RawReference {
                         callee,
@@ -346,6 +362,7 @@ export function App() {
         <div className="wrap">
             <Hello name="world" />
             <Panel.Header title="hi"></Panel.Header>
+            <Icons.arrow />
         </div>
     );
 }
@@ -359,6 +376,11 @@ export function App() {
         assert!(
             callees.contains(&"Header"),
             "dotted JSX element should be a call site: {callees:?}"
+        );
+        assert!(
+            callees.contains(&"arrow"),
+            "a dotted tag is a value in scope whatever the case of its \
+             property, so <Icons.arrow /> is a call site: {callees:?}"
         );
         assert!(
             !callees.contains(&"div"),
--- a/crates/vera-core/src/parsing/references.rs
+++ b/crates/vera-core/src/parsing/references.rs
@@ -51,7 +51,7 @@ fn is_call_node(lang: Language, kind: &str) -> bool {
     match lang {
         Language::Rust => matches!(kind, "call_expression" | "macro_invocation"),
         Language::TypeScript | Language::JavaScript => {
-            matches!(kind, "call_expression" | "new_expression")
+            matches!(kind, "call_expression" | "new_expression") || is_jsx_element_node(kind)
         }
         Language::Python => kind == "call",
         Language::Go => kind == "call_expression",
@@ -83,6 +83,24 @@ fn is_call_node(lang: Language, kind: &str) -> bool {
         Language::Elm => kind == "function_call_expr",
         _ => false,
     }
+}
+
+/// Whether a node kind is a JSX element tag that names the component it renders.
+///
+/// `jsx_closing_element` is deliberately excluded so `<Foo></Foo>` records one
+/// reference rather than two.
+fn is_jsx_element_node(kind: &str) -> bool {
+    matches!(kind, "jsx_opening_element" | "jsx_self_closing_element")
+}
+
+/// Whether a JSX tag names a host element rather than a component in scope.
+///
+/// JSX resolves a tag beginning with a lowercase letter, such as `<div>`, to an
+/// intrinsic element string, and any other, such as `<Foo>` or `<Foo.Bar>`, to a
+/// value in scope. Only the latter is a call site; recording the former would
+/// add an edge per HTML tag in the repository.
+fn is_jsx_host_element(kind: &str, name: &str) -> bool {
+    is_jsx_element_node(kind) && name.starts_with(|ch: char| ch.is_lowercase())
 }
 
 /// Extract the callee name from a call node.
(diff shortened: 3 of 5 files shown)


citron07r:
Thanks — both parts addressed, though the second one differently than suggested. Pushed as `47bde6b`.

## `.tsx` structural-search test: added

`structural_search_works_in_jsx_bearing_tsx` in `crates/vera-core/src/retrieval/structural.rs`
covers a JSX-bearing `.tsx` file: a real `process.env.API_URL` read is found and attributed to the
enclosing `Panel`, while a commented-out decoy and a `process.env.FAKE_URL` string sitting in a JSX
attribute are filtered out. Nothing covered `.tsx` structural search before, so this is new
coverage either way.

**It is a characterisation test, not a regression test, and the doc comment says so.** I checked it
against the reverted change, per this repo's rule that a new test has to be shown to catch the bug:
it passes with the grammar switch reverted, both for `SyntaxFilter::new` and for the indexing-side
`parse_file_with_diagnostics`. Two reasons:

- tree-sitter still lexes comments and string literals inside an error region, so both decoys are
  rejected under the `typescript` grammar too
- `export function Panel() { … }` is still recognised as a `function_declaration` under the
  `typescript` grammar; only the JSX in its body errors, so the match is attributed either way

I could not construct any structural-search input where the grammar choice changes the result.

## Full Semble suite: not run, because this change cannot reach the corpus

The `retrieval/**` path instruction is the right default, but it does not apply here on the
evidence:

**1. The `SyntaxFilter` change is behaviour-neutral**, per the reverted-test result above. It picks a
different grammar to answer the same question — "is this byte range inside a comment or string" —
and both grammars answer it identically, because the lexer handles comments and strings before the
parser decides anything about JSX.

**2. The corpus contains no `.tsx` under any benchmarked root.** `eval/semble-subset-corpus.toml`
pins a `benchmark_root` per repo, and only three of the sixteen repos are JS/TS:

| repo | `benchmark_root` | `.tsx` in root | `.tsx` in repo |
|---|---|---|---|
| `express` | `lib` | 0 | 0 |
| `axios` | `lib` | 0 | 0 |
| `zod` | `packages/zod/src` | **0** | 29 |

zod is the only one that ships `.tsx` at all, and all 29 files are under `packages/docs/`
(19 in `components/`, 9 in `app/`, 1 in `pages/`), outside the benchmarked `packages/zod/src`. So
no file whose parse this PR changes is indexed by the suite, and a before/after run would compare
two identical indexes.

Happy to run it anyway if you would rather have the number on record — say the word and I will set
up the harness. I would rather tell you it cannot move than hand you a run whose result is
predetermined.

## Note on CI

No build or test check appears on this PR — only `cubic`, `[code]smith` and `CodeRabbit`. If the
Rust workflow is meant to run on `pull_request` from forks, it may need approval. Locally on this
branch: `cargo build --release` clean for the workspace, `cargo test --workspace` 921 passed / 0
failed, `cargo fmt --check` clean, and `cargo clippy -p vera-core --lib` reporting only the five
warnings already present on `master`.

citron07r:
@coderabbitai review

citron07r:
@cubic-dev-ai good catch, that was a real bug and it is fixed in `c7a52ee`.

`<Icons.arrow />` was being dropped. The host-element check ran on the *extracted callee*, and
`extract_callee` returns the rightmost identifier of a member expression, so the tag produced the
callee `arrow`, the lowercase rule read that as an intrinsic element, and a genuine component
reference disappeared. Exactly the failure mode this PR exists to fix, reintroduced through a
different door.

Host status now comes from the tag node instead of the name string:

- `name` field is a `member_expression` → never a host element, whatever the case of the property,
  because a dotted tag is a member access on a value in scope
- otherwise → host element only if the tag text starts with a lowercase letter

Reproduced first, then fixed. Added `<Icons.arrow />` to `tsx_jsx_elements_are_call_sites`, which
failed with `["Hello", "Header"]` before the change. Verified end to end on an indexed fixture:

```
Hello    -> app.tsx:4-8 function:App
arrow    -> app.tsx:5-9 function:App
div      -> No callers found for 'div'.
circle   -> No callers found for 'circle'.
```

`<div>` and the namespaced `<svg:circle>` still stay out of the reference graph.

One thing left as-is, flagging rather than deciding silently: the recorded callee for `<Icons.arrow />`
is `arrow`, not `Icons`. That follows the existing convention for member expressions, where
`obj.method()` records `method`. Whether a dotted JSX tag should instead record the root value is a
separate question about `extract_callee` that would change method-call behaviour too, so I have not
touched it here.

`cargo test --workspace` 921 passed / 0 failed, `cargo fmt --check` clean, clippy unchanged from
`master`.

citron07r:
@coderabbitai full review

citron07r:
Ran the CodeRabbit CLI locally against `master` while the hosted review quota recovered. One finding,
valid, fixed in `7b7364f`.

## Intrinsic tags are ASCII-only

`char::is_lowercase` is Unicode-aware, so `<éWidget />` was classified as a host element and its
reference dropped. The transpilers test ASCII specifically — Babel's `isCompatTag` is `/^[a-z]/`,
TypeScript's `isIntrinsicJsxName` compares against `a`..`z` — so a tag starting outside that range is
an identifier in scope and therefore a component reference.

Now `ch.is_ascii_lowercase()`. Confirmed the test catches the regression: restoring `is_lowercase`
gives `["Hello", "Header", "arrow"]`, missing `éWidget`.

`tsx_jsx_elements_are_call_sites` now pins the whole classification in one place: `<Hello />` and
`<Panel.Header>` and `<Icons.arrow />` and `<éWidget />` are call sites; `<div>` is not.

## Benchmark evidence is coming after all

I said earlier that the Semble subset contains no `.tsx` under any benchmark root, which is true. But
the Vera eval suite in `eval/tasks/` runs against a different corpus, and that one is relevant:

```
repo         .tsx    .ts    .js
fastify         0     35    255
flask           0      0      0
ripgrep         0      0      0
turborepo     303    506    236
```

turborepo carries 303 `.tsx` files, so this PR does change what gets indexed there. I am running
`benchmarks/scripts/run_local_binary_benchmarks.py` over the 21-task suite with released v1.0.0 as
the control and this branch as the candidate, and will post the before/after table with per-task
diffs.

Correcting myself on the earlier comment: "cannot reach the corpus" was accurate for the Semble
subset and too broad as a general claim. The eval corpus does reach it.

`cargo test --workspace` 921 passed / 0 failed, `cargo fmt --check` clean, clippy unchanged from
`master`.

citron07r:
@cubic-dev-ai correct again, fixed in `45e561a`.

`<My-element />` was being recorded as a component reference. Checked against the current
TypeScript source rather than from memory, since this predicate has now bitten me three times:

```ts
export function isIntrinsicJsxName(name: __String | string): boolean {
    const ch = (name as string).charCodeAt(0);
    return (ch >= CharacterCodes.a && ch <= CharacterCodes.z) || (name as string).includes("-");
}
```

So a hyphen makes a tag intrinsic whatever its case — the custom-element form — and the compiler
emits it as the string `"My-element"` rather than a value lookup. Indexing those would add an edge
per custom element, which is the same noise problem as `<div>`.

Worth noting for anyone reading this later: the `:` clause this function used to carry is **not** in
current TypeScript, so namespaced tags are no longer special-cased. `<svg:circle />` still stays out
because it starts with a lowercase ASCII letter, not because of a colon rule.

The predicate now reads as one rule with the reason attached, and
`tsx_jsx_elements_are_call_sites` pins every branch of it in one place:

| tag | recorded | why |
|---|---|---|
| `<Hello />` | yes | bare identifier, not `a`..`z` |
| `<Panel.Header>` | yes | member expression |
| `<Icons.arrow />` | yes | member expression, lowercase property is irrelevant |
| `<éWidget />` | yes | `é` is outside ASCII `a`..`z` |
| `<div>` | no | intrinsic |
| `<My-element />` | no | hyphen, intrinsic |

Confirmed the test catches the regression: dropping the hyphen clause puts `My-element` back in the
callee list.

`cargo test --workspace` 921 passed / 0 failed, `cargo fmt --check` clean, clippy unchanged from
`master`.

PR #63, opened 2026-08-19
fix(upgrade): verify the version changed, not that it equals the release tag

### LINKED ISSUE #62: vera upgrade --apply reports a successful upgrade as failure when the registry resolves a version other than the GitHub tag
`vera upgrade --apply` verifies that an upgrade took effect by comparing the version the
installer recorded against the version derived from the GitHub release tag:

```rust
// crates/vera-cli/src/commands/upgrade.rs
fn applied_version_mismatch(method: &str, latest: Option<&str>, installed: Option<&str>) -> Option<String> {
    let (latest, installed) = (latest?, installed?);
    if installed == latest {
        return None;
    }
    ...
}
```

Those two values come from different sources. `latest` is read from the GitHub release, while
`installed` is whatever the package manager actually resolved. When they disagree for any reason
other than a stale registry, a **successful** upgrade is reported as a failure, the command exits
non-zero, and the hint tells the user to install the GitHub-tag version.

## Impact

If the registry is *ahead* of the GitHub release, the hint advises a downgrade.

| current | GitHub `latest` | installer resolved | reported | actual |
|---|---|---|---|---|
| 1.0.0 | 1.0.1 | 1.0.1 | success | correct |
| 1.0.0 | 1.0.1 | **1.0.2** | **failure**, "install 1.0.1" | upgraded fine, hint is a downgrade |
| 1.0.0 | 1.2.0 | **1.1.0** | **failure**, "install 1.2.0" | upgraded fine |
| 1.0.0 | 1.0.1 | 1.0.0 | failure | correct, this is the real no-op |

Only the last row is the condition the check exists to catch. The npm and PyPI wrappers resolve
independently of the GitHub release, so rows 2 and 3 are reachable in normal operation, and this
repository has already seen the two registries drift apart (#35).

## Cause

The success condition asks "does the installed version equal the advertised one" when the question
it needs to answer is "did the installed version change". The no-op being guarded against is
precisely "the installer resolved the version that was already installed", which is a comparison
against the *current* version, not against `latest`.

## Suggested fix

Compare `installed` against `current_version`. That detects the no-op exactly, accepts any genuine
version change regardless of which source resolved it, and no longer requires `latest` to be known
in order to verify at all.

Reported originally as review feedback on #43 by cubic; filing it so it is tracked rather than lost
in a merged PR's thread.

citron07r:
Fixes #62.

Follow-up to review feedback left on #43 by @cubic-dev-ai that was never actioned before that PR
merged. The observation was correct.

## What was wrong

`applied_version_mismatch` treated an upgrade as applied only when the installer-recorded version
equalled the GitHub release tag:

```rust
if installed == latest { return None; }   // success
```

`latest` comes from the GitHub release; `installed` is whatever the package manager resolved. They
are independent sources, so they can disagree without anything being broken. When they did, a
successful upgrade was reported as a failure, `vera upgrade --apply` exited non-zero, and the hint
told the user to install the GitHub-tag build — a downgrade whenever the registry was ahead.

| current | GitHub `latest` | resolved | before | after |
|---|---|---|---|---|
| 1.0.0 | 1.0.1 | 1.0.1 | pass | pass |
| 1.0.0 | 1.0.1 | 1.0.2 | **fail, hints downgrade** | pass |
| 1.0.0 | 1.2.0 | 1.1.0 | **fail** | pass |
| 1.0.0 | 1.0.1 | 1.0.0 | fail | fail (correct: the real no-op) |

Only the last row is what the guard exists to catch, and this repo has already seen npm and PyPI
drift from the GitHub release (#35), so the middle rows are reachable in normal operation.

## What changed

The check now asks whether the installed version *changed*, comparing against `current_version`.
That detects the no-op exactly, accepts a genuine upgrade whichever source resolved it, and no
longer requires `latest` to be known in order to verify at all — the message falls back to "the new
version" when it is absent.

The failure message changed from `expected {latest}, but the installer recorded {installed}` to
`still on {installed}`, which is what actually happened in the no-op case.

## Verification

`cargo test --workspace` 917 passed / 0 failed. `cargo fmt --check` clean.
`cargo clippy -p vera-cli` reports only the five warnings already present on `master`.

Four new tests cover the rows above, including the two that previously failed:

- `accepts_a_version_the_registry_resolved_ahead_of_the_release`
- `accepts_an_intermediate_version_between_current_and_latest`
- `detects_the_no_op_even_when_the_release_version_is_unknown`
- `keeps_a_no_op_upgrade_unapplied`

The existing no-op coverage from #43 is kept and still passes, so the behaviour that PR added is
preserved; only the false failures are removed.

---
## Summary by cubic
Verifies upgrades by checking that the installer’s recorded version changes across the install instead of matching the GitHub release tag. Only trusts a baseline provenance that identifies the running binary and flags downgrades.

- Compares the post‑install recorded version to pre‑install provenance tied to the current binary; change = success, unchanged = failure with “still on {installed}”.
- Uses `semver` (v‑prefix tolerant) to order versions; downgrades (installed < baseline) are failures; unparseable versions yield Unknown.
- Returns Unknown when baseline or post‑install provenance is missing or the baseline does not identify the running binary; `latest` only words the hint (fallback: “new version”); cleans hint text and keeps the Releases fallback.
- Adds tests for registry‑ahead, intermediate versions, downgrades, baseline validation vs running binary, unknown‑`latest` hint wording, and the no‑op; moves the test module to the file bottom.

<sup>Written for commit 92d5a97042d499abc20fc7061637260ff3e28536. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/63?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved upgrade verification by comparing the installed version with a validated pre-upgrade baseline.
  - Recognizes successful upgrades when the installed version changes without being a downgrade.
  - Correctly handles no-op upgrades, unavailable release information, and versions with a leading “v”.
  - Reports missing or unparseable version information as unknown instead of incorrectly indicating success or failure.
  - Rejects downgrade results and improves release-version guidance.

- **Tests**
  - Expanded coverage for version validation, ordering, no-ops, downgrades, and unknown versions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4028,6 +4028,7 @@ dependencies = [
  "console 0.16.4",
  "dirs",
  "reqwest",
+ "semver",
  "serde",
  "serde_json",
  "tempfile",
--- a/crates/vera-cli/Cargo.toml
+++ b/crates/vera-cli/Cargo.toml
@@ -17,6 +17,7 @@ vera-serve = { workspace = true }
 clap = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+semver = "1"
 tokio = { workspace = true }
 anyhow = { workspace = true }
 reqwest = { workspace = true }
--- a/crates/vera-cli/src/commands/upgrade.rs
+++ b/crates/vera-cli/src/commands/upgrade.rs
@@ -56,12 +56,22 @@ pub fn run(apply: bool, json_output: bool) -> Result<()> {
         .install_method
         .as_deref()
         .expect("apply requires a resolved install method");
+
+    // Read what the installer recorded *before* running it. The no-op is
+    // "the installer resolved the same version again", so the comparison has to
+    // be against the installer's own previous record. Comparing against the
+    // running binary's version instead would misread a stale or foreign
+    // provenance record as a successful upgrade.
+    let baseline_version = state::load_install_provenance()
+        .ok()
+        .and_then(|provenance| provenance.version);
+
     update_check::apply_update(method)?;
 
     // The installer command exiting 0 does not mean the new version landed. A
     // package registry can lag behind a GitHub release, in which case the
     // installer resolves the version already installed and the upgrade
-    // silently no-ops. Check whether the installed version actually changed
+    // silently no-ops. Check whether the recorded version actually changed
     // rather than reporting success on the exit code alone.
     report.installed_version = state::load_install_provenance()
         .ok()
@@ -69,7 +79,7 @@ pub fn run(apply: bool, json_output: bool) -> Result<()> {
 
     match verification_outcome(
         method,
-        &report.current_version,
+        baseline_version.as_deref(),
         report.latest_version.as_deref(),
         report.installed_version.as_deref(),
     ) {
@@ -97,22 +107,30 @@ pub fn run(apply: bool, json_output: bool) -> Result<()> {
 /// installer keeps resolving the version already installed, every installer
 /// command exits 0, and the upgrade silently no-ops.
 ///
-/// The test is whether the installed version *changed*, not whether it equals
-/// the advertised one. The installer resolves from the package registry while
-/// `latest` comes from the GitHub release, and the two can legitimately differ:
-/// a registry that is ahead of the release, or one that resolves an
-/// intermediate version, still performed a real upgrade. Comparing against
-/// `latest` reports those as failures and points the user at an older build.
+/// The test is whether the installer's own record *changed*, comparing the
+/// provenance written before the install against the one written after.
+///
+/// Not against `latest`: that comes from the GitHub release while the installer
+/// resolves from a package registry, and the two can legitimately differ. A
+/// registry ahead of the release, or one resolving an intermediate version,
+/// still performed a real upgrade, and comparing against `latest` would report
+/// those as failures and point the user at an older build.
+///
+/// Not against the running binary's version either: provenance can be stale or
+/// belong to another installation, so a no-op could leave it differing from the
+/// running version and read as success.
+///
+/// `latest` is used only to word the hint.
 fn applied_version_mismatch(
     method: &str,
-    current: &str,
+    baseline: &str,
     latest: Option<&str>,
     installed: &str,
 ) -> Option<String> {
-    if installed != current {
+    if installed != baseline {
         return None;
     }
-    let target = latest.unwrap_or("the new version");
+    let target = latest.unwrap_or("new version");
     Some(format!(
         "upgrade did not take effect: still on {installed}.\n\
          This usually means the {method} package for {target} has not been published yet, so the \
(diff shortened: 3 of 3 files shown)


citron07r:
Thanks both — four findings, all valid, fixed in `bbf0e4a`. One of them showed my fix was still
wrong in the same family as the original bug.

## The baseline was still the wrong thing to compare against

@cubic-dev-ai's P1 is the substantive one. Comparing post-install provenance against the **running
binary's version** is better than comparing against the release tag, but still not the question the
check needs to answer. Provenance can be stale or belong to another installation, so a no-op can
leave it differing from the running version and be read as a successful upgrade.

The baseline is now the provenance record read **before** `apply_update`:

```rust
let baseline_version = state::load_install_provenance().ok().and_then(|p| p.version);
update_check::apply_update(method)?;
report.installed_version = state::load_install_provenance().ok().and_then(|p| p.version);
```

"Did the installer change its own record" is exactly what a no-op is, and it no longer depends on
provenance agreeing with the running binary. Missing either side yields `Unknown` rather than an
assumed outcome, as suggested.

Pinned by `detects_the_no_op_when_provenance_disagrees_with_the_running_binary`, which fails under
the version I pushed yesterday.

## "the the new version"

Real, and mine. The template says `install the {target} binary` and the fallback was
`"the new version"`. Fallback is now `"new version"`, and
`renders_a_readable_hint_when_the_release_version_is_unknown` asserts the rendered string contains
no `"the the"`.

## The unknown-release path is unreachable — agreed

You are both right that `update_available()` is `false` when `latest_version` is `None`, so `run`
returns before `apply_update` and verification never sees that case. I have not changed the apply
gate: letting `--apply` proceed without knowing the target version is a behaviour change that wants
its own discussion.

What I have done is stop implying otherwise. With the baseline fix, `latest` is used **only to word
the hint**, never for detection, so the `unwrap_or` is wording rather than a correctness fallback,
and the doc comment says so. The claim in my original PR description that it "no longer requires
`latest` to be known in order to verify" was overstated for the code path that actually runs.

## Test module placement

Moved to the file bottom per the Rust path instruction. Worth noting it was already above
`print_report` and `install_method_source_name` before this branch — adding tests to it made an
existing misplacement bigger, so fixing it here seemed right rather than leaving it worse than I
found it.

## Verification

`cargo test --workspace` 917 passed / 0 failed. `cargo fmt --check` clean.
`cargo clippy -p vera-cli` reports only the five warnings already on `master`.

PR #65, opened 2026-08-19
fix(retrieval): bound the KNN candidate pool instead of clamping silently

### LINKED ISSUE #64: KNN candidate pool is silently truncated at the sqlite-vec cap, so the filter and metadata over-fetch stop working
`VectorStore::search` clamps the KNN `k` to `MAX_KNN_K` (4096), which is the sqlite-vec hard limit.
That clamp is correct and is what stopped vector search failing outright (#38 / #42). The problem is
that nothing upstream knows it exists, so the candidate pool is computed as if it were unbounded and
the truncation is invisible.

## How the pool is built

Three multipliers compound before the request reaches storage:

| step | location | effect |
|---|---|---|
| query-type scaling | `compute_vector_candidates(limit, multiplier)` in `retrieval/hybrid.rs` | `limit * multiplier`, min 50 |
| filter over-fetch | `run_hybrid_search` in `retrieval/hybrid.rs` | `* 4`, min 200, when filters are active |
| metadata over-fetch | `search_vector_with_stores` in `retrieval/vector.rs` | `* 1.5`, min `limit + 10` |
| clamp | `VectorStore::search` in `storage/vector.rs` | `min(4096)`, logged at `debug` only |

A natural-language query with an active filter therefore asks for `limit * 5 * 4 * 1.5`. At
`limit = 20` that is 600; at `limit = 200` it is 6000, past the cap.

## Why it matters

The clamp is silent. It is emitted at `tracing::debug!`, which is off by default, and nothing in the
returned value tells the caller the pool was reduced. Two consequences:

1. **The over-fetch stops working exactly when it is needed.** The `* 4` filter over-fetch and the
   `* 1.5` metadata over-fetch exist so that post-fetch filtering and missing metadata do not empty
   the window. Once the request is clamped, those multipliers are inert: a selective filter over a
   4096-row window can still under-fill the vector arm, weakening RRF and giving the reranker less
   to work with.
2. **It is a silent behaviour change in a public API.** `VectorStore::search` is `pub`. A caller
   asking for more than 4096 gets fewer results with no error and no signal.

This is strictly better than the previous behaviour, where exceeding the cap was a hard error that
dropped the whole vector arm and fell back to BM25-only. It is not a regression. But the same
lesson from #38 applies: the reason that bug was undiagnosable was a swallowed signal, and this is a
quieter version of the same shape.

## Possible directions

- Bound the pool where it is computed, so the arithmetic in `hybrid.rs` and `retrieval/vector.rs`
  is honest about the ceiling rather than relying on the storage layer to silently fix it up.
- Raise the clamp log from `debug` to `warn` so `-v` shows it.
- Return the requested-vs-fetched count from `VectorStore::search` so callers can decide.

The first two change no ranking behaviour: the same `k` still reaches sqlite-vec, so the same
candidates come back. The third is an API change and would want maintainer input.

Reported originally as review feedback on #42 by cubic; filing it so it is tracked rather than lost
in a merged PR's thread.

citron07r:
Fixes #64.

Follow-up to review feedback left on #42 by @cubic-dev-ai that was never actioned before that PR
merged. The observation was correct.

## What was wrong

The clamp in `VectorStore::search` is what keeps vector search working at all — before #42, exceeding
the cap was a hard error that dropped the entire vector arm and fell back to BM25-only (#38). That
part is right and stays.

The problem is that nothing upstream knew the cap existed. Three multipliers compound before the
request arrives:

| step | effect |
|---|---|
| `compute_vector_candidates` | `limit * multiplier`, min 50 |
| filter over-fetch in `run_hybrid_search` | `* 4`, min 200, when filters are active |
| metadata over-fetch in `search_vector_with_stores` | `* 1.5`, min `limit + 10` |
| `VectorStore::search` | `min(4096)`, logged at `debug` |

So the pool was computed as if unbounded and the shortfall was invisible: `debug` is off by default
and nothing in the return value signals it. Two consequences:

1. **The over-fetch goes inert exactly when it is needed.** Both over-fetches exist so that
   post-fetch filtering and missing metadata cannot empty the window. Above the cap they do nothing,
   so a selective filter over a 4096-row window can still under-fill the vector arm feeding RRF and
   the reranker.
2. **Silent behaviour change in a public API.** `VectorStore::search` is `pub`; a caller asking for
   more than 4096 gets fewer results with no error.

## What changed

- `MAX_KNN_K` is now `pub`, so callers can size against the real ceiling.
- `candidate_pool` sizes the request against that ceiling and returns `(requested, fetched)`, and the
  caller warns when they differ, naming the query.
- The storage clamp stays as a backstop, with its log raised from `debug` to `warn` — reaching it now
  means an external caller, not Vera's own path.
- `limit + 10` became `saturating_add(10)`; it could overflow on a caller-supplied limit near
  `usize::MAX`.

## On the retrieval path instruction

**No ranking change, and no benchmark run.** The same `k` reaches sqlite-vec before and after, so the
same candidates come back in the same order; only the reporting differs. The pool arithmetic is
moved, not altered:

- below the cap, `candidate_pool` returns exactly what the old expression did — pinned by
  `assert_eq!(candidate_pool(10), (20, 20))` and `assert_eq!(candidate_pool(100), (150, 150))`
- above it, the value passed to `search` is `MAX_KNN_K`, which is what the storage clamp produced
  anyway

Happy to run the suite if you would rather have it on record, but I did not want to present a run
whose result is determined by construction.

## Verification

`cargo test --workspace` 916 passed / 0 failed. `cargo fmt --check` clean.
`cargo clippy -p vera-core --lib` reports only the five warnings already present on `master`.

`candidate_pool_is_bounded_by_the_knn_cap` is a pure test with no assets, covering the below-cap,
at-cap and `usize::MAX` cases.

---
## Summary by cubic
Bounds the KNN candidate pool to the sqlite-vec cap and surfaces shortfalls without logging user queries. Previously, `VectorStore::search` silently clamped k>4096 at `debug`; now retrieval sizes against the cap, warns on short pools, and ranking is unchanged.

- Exposes `MAX_KNN_K` and adds `candidate_pool(limit) -> (requested, fetched)`; emits `warn` when `requested > fetched` (warning omits the query text).
- Keeps the storage-level clamp as a backstop and raises its log to `warn`.
- Switches `limit + 10` to `saturating_add(10)` to avoid overflow.
- Migration: external callers must size candidate pools to `MAX_KNN_K` to avoid warnings and inert over-fetch.

<sup>Written for commit ef9bd4c5337579be03a6aa41a81ed863d89c7a0f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/65?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/vera-core/src/retrieval/vector.rs
+++ b/crates/vera-core/src/retrieval/vector.rs
@@ -62,8 +62,10 @@ pub async fn search_vector_with_stores(
     // 2. Search the vector store for nearest neighbors.
     let (requested, candidates) = candidate_pool(limit);
     if requested > candidates {
+        // Deliberately without the query text: the surrounding `debug!` calls
+        // carry it, but this one is on by default, and a search query is user
+        // content that should not be raised into default-visible logs.
         warn!(
-            query = query,
             requested,
             fetched = candidates,
             "candidate pool exceeds the sqlite-vec KNN cap; the filter and \
--- a/crates/vera-core/src/retrieval/vector.rs
+++ b/crates/vera-core/src/retrieval/vector.rs
@@ -7,11 +7,11 @@
 //! don't appear literally in results (e.g., "memory allocation" finds `alloc`).
 
 use anyhow::Result;
-use tracing::debug;
+use tracing::{debug, warn};
 
 use crate::embedding::{EmbeddingError, EmbeddingProvider};
 use crate::storage::metadata::MetadataStore;
-use crate::storage::vector::VectorStore;
+use crate::storage::vector::{MAX_KNN_K, VectorStore};
 use crate::types::SearchResult;
 
 /// Errors specific to vector search.
@@ -60,9 +60,16 @@ pub async fn search_vector_with_stores(
     );
 
     // 2. Search the vector store for nearest neighbors.
-    // Fetch extra candidates to account for missing metadata without doubling
-    // large caller-selected candidate pools.
-    let candidates = limit.saturating_add(limit / 2).max(limit + 10);
+    let (requested, candidates) = candidate_pool(limit);
+    if requested > candidates {
+        warn!(
+            query = query,
+            requested,
+            fetched = candidates,
+            "candidate pool exceeds the sqlite-vec KNN cap; the filter and \
+             metadata over-fetch are inert above it"
+        );
+    }
 
     let vector_results = vector_store
         .search(&query_embedding, candidates)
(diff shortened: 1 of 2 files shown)


citron07r:
@coderabbitai review

Issue #66, opened 2026-08-19
Default local embedding and reranker models are CC-BY-NC-4.0, so the default setup is non-commercial only


citron07r:
Both models Vera downloads by default for the local backend are licensed **CC-BY-NC-4.0**, which
permits research and evaluation only:

| constant | repo | licence |
|---|---|---|
| `EMBEDDING_REPO` | `jinaai/jina-embeddings-v5-text-nano-retrieval` | `cc-by-nc-4.0` |
| `RERANKER_REPO` | `jinaai/jina-reranker-v2-base-multilingual` | `cc-by-nc-4.0` |

Both taken from the `license:` tag on the Hugging Face API for each repo. The reranker card states
it directly:

> This model repository is licenced for research and evaluation purposes under CC-BY-NC-4.0. For
> commercial usage, please refer to Jina AI's APIs, AWS Sagemaker or Azure Marketplace offerings.

## Why this matters

Vera itself is open source, and `vera setup` fetches these automatically as the default local
backend. Anyone running Vera to search a commercial codebase is using a non-commercially-licensed
model without being told. Nothing in `vera setup`, `vera doctor` or the README surfaces the licence.

`Zenabius/CodeRankEmbed-onnx`, offered as the `coderankembed` option, declares **no licence at
all**, and has 5 downloads. A re-export with no declared licence is a separate ambiguity.

## Permissively licensed alternatives that fit the existing loader

Both already ship the exact file layout `local_models/mod.rs` expects
(`onnx/model_quantized.onnx`, `onnx/model_fp16.onnx`, `tokenizer.json`):

| role | candidate | licence | dim | context | notes |
|---|---|---|---|---|---|
| embedding | `jinaai/jina-embeddings-v2-base-code` | **apache-2.0** | **768** | 8192 | English + 30 programming languages, 394k downloads, official Jina repo |
| reranker | `BAAI/bge-reranker-v2-m3` via `onnx-community/bge-reranker-v2-m3-ONNX` | apache-2.0 upstream | — | — | 24.5k downloads on the ONNX export |

`jina-embeddings-v2-base-code` reports `hidden_size: 768` in its `config.json`, the same as the
current `EMBEDDING_DIM`, so the sqlite-vec schema is unchanged; only a reindex is needed.

The bge reranker caveat: `BAAI/bge-reranker-v2-m3` is `apache-2.0`, but the popular
`onnx-community` re-export declares no licence of its own. Exporting ONNX from the Apache-2.0
[...]

PR #72, opened 2026-08-19
feat(local-models): make CodeRankEmbed the default embedding model

### LINKED ISSUE #70: Default embedding is a general-text model with no published code-retrieval numbers; the code-specialised option Vera already ships scores +10.7 CSN over the next best
Vera is a code-search tool, but its default local embedding model is a **general-text** model, and
the **code-specialised** model Vera already ships as an option has published numbers showing it is
substantially better at code retrieval.

## Published evidence

`nomic-ai/CodeRankEmbed`'s model card publishes a direct comparison. CSN is CodeSearchNet MRR, CoIR
is NDCG@10:

| model | params | CSN (MRR) | CoIR (NDCG@10) |
|---|---|---|---|
| **CodeRankEmbed** | 137M | **77.9** | **60.1** |
| `jina-embeddings-v2-base-code` | 161M | 67.2 | 58.4 |
| CodeSage-Base | 356M | 68.7 | 57.5 |
| Arctic-Embed-M-Long (CodeRankEmbed's general-purpose base) | 137M | 53.4 | 43.0 |

Source: https://huggingface.co/nomic-ai/CodeRankEmbed

The last row is the useful control: same 137M base model, code fine-tuning is the only difference.
**CSN 53.4 → 77.9 (+24.5) and CoIR 43.0 → 60.1 (+17.1).** Code specialisation is worth roughly 15-25
points on code-retrieval metrics at equal size. That is not a marginal effect.

## The current default has no code-retrieval evidence at all

`jina-embeddings-v5-text-nano-retrieval` (239M) publishes general MTEB retrieval (63.3). Neither its
card, nor jina.ai's model page, nor the `-small-retrieval` card publishes a CoIR or CodeSearchNet
number. Strong general-text MTEB placement should not be assumed to transfer to code retrieval —
the Arctic-Embed row above is exactly a case where a capable general model scores far below a
code-tuned one of the same size.

So the comparison that would justify the current default does not exist publicly.

## Why this is cheap to act on

Vera **already supports CodeRankEmbed**. `local_models/mod.rs` has `CODERANK_EMBEDDING_REPO` and
`CODERANK_QUERY_PREFIX` ("Represent this query for searching relevant code:"), and
`LocalEmbeddingModelConfig::coderankembed()` exists alongside `jina()`. It is a supported
non-default option, wired end to end.

It is also **smaller** than the current default: 137M against 239M.

## The gap is packaging, again

`nomic-ai/CodeRankEmbed` is **MIT** with 419k downloads, but ships no ONNX. Vera points at
`Zenabius/CodeRankEmbed-onnx`, which has **5 downloads and no declared licence** (#66). So the
better-evidenced model currently rides on the weakest supply chain in the project.

An in-repo ONNX export from the MIT upstream would fix both the provenance problem and the
dependency-thinness problem in one step.

## Suggested outcome

1. Consider making CodeRankEmbed the default for a code-search tool, on the published numbers above.
2. Either way, replace the 5-download re-export with a trusted or self-produced ONNX artifact.
3. If the v5 line is kept as default, it would be worth publishing a code-retrieval comparison to
   justify it, because none currently exists.

## What is not established

I have not benchmarked either model on Vera's own eval corpus. These are the model authors'
published numbers on standard benchmarks, which is stronger evidence than the parameter counts and
release dates I was reasoning from earlier, but it is not a measurement of Vera's actual retrieval
quality. An A/B on `eval/tasks/` would settle it, and the dimension difference matters for that:
CodeRankEmbed would need its own reindex.

### LINKED ISSUE #71: CodeRankEmbed cannot index on the CoreML backend: inference fails with an inconsistent static output shape (same class as #41)
`CodeRankEmbed`, offered as the `coderankembed` embedding option, cannot index at all on the CoreML
backend. Reproduced twice on Apple Silicon with the released v1.0.0 binary:

```
[E:onnxruntime:, sequential_executor.cc:572 ExecuteKernel] Non-zero status code returned while
running 9868062452848548768_CoreML_9868062452848548768_1 node.
Name:'CoreMLExecutionProvider_9868062452848548768_CoreML_9868062452848548768_1_1'
Status Message: Exception: .../coreml/model/model.mm:79 GetStaticOutputShape(...)
inferred_dim == -1 || inferred_dim == coreml_static_dim was false.
CoreML static output shape ({2,1,1,80}) and inferred shape ({1,1,1,-1}) have an inconsistent
static dimensions (2 vs. 1).

Error: indexing failed: embedding generation failed: embedding API error (status 500)
```

Two source files and a two-line fixture are enough to trigger it.

## Same class as #41

This is the failure mode #41/#33 already documented for the reranker: the CoreML EP accepts a fused
subgraph at session construction, so a probe that only builds a session passes, and then inference
fails. Here it surfaces on the **embedding** side instead, and it is fatal rather than silent —
indexing aborts.

The existing fix shape is `reranker_execution_provider` in `local_models/mod.rs:65-72`, which maps
CoreML to CPU for the reranker. There is no equivalent for the embedding path.

## It works on CPU

```
$ vera index . --onnx-jina-cpu
Indexing complete!   Chunks created: 2
```

Search over that index also returns vector results correctly on a build from current `master`
(stored dim is `float[768]`, matching `EMBEDDING_DIM`).

Note for anyone reproducing: the released v1.0.0 binary reports the follow-up search failure as a
bare `storage error: vector search failed`, because it predates #42's `{0:#}` error-chain fix. Use a
current build or the cause is invisible.

## Why this matters more than an option being broken

`CodeRankEmbed` is the strongest code-retrieval model Vera ships by published numbers — CodeSearchNet
MRR 77.9 and CoIR NDCG@10 60.1, against 67.2 / 58.4 for `jina-embeddings-v2-base-code` and 53.4 /
43.0 for its own general-purpose base (https://huggingface.co/nomic-ai/CodeRankEmbed). See #70.

So the best-evidenced option for a code-search tool is unusable on what is likely the most common
developer setup for this project.

## The tradeoff that needs deciding

Pinning the embedding to CPU under CoreML, mirroring the reranker fix, would make it work
everywhere. But CoreML currently accelerates **embeddings only** — roughly 39 ms against the
reranker's seconds — so pinning would give up the one thing CoreML does well here and slow indexing
materially on Apple Silicon.

That is a real quality-versus-indexing-speed decision rather than a clear bug fix, which is why this
is an issue and not a PR. Options as I see them:

1. Pin embeddings to CPU when the model is CodeRankEmbed and the EP is CoreML, and document the
   indexing cost.
2. Fix or re-export the ONNX so CoreML can execute it — the error is a static/dynamic output-shape
   disagreement, which may be an artifact of how `Zenabius/CodeRankEmbed-onnx` was exported. That
   re-export has 5 downloads and no declared licence (#66), so a fresh export from the MIT upstream
   may resolve this and the provenance problem together.
3. Detect the failure and fall back with a warning instead of aborting the index.

## Not established

I have not completed an A/B of CodeRankEmbed against the current default on Vera's own eval corpus —
the run aborted on exactly this bug. The nano baseline is measured (recall@1 0.704, recall@5 0.898,
recall@10 0.926, MRR 0.861, nDCG 0.859 over 9 ripgrep+fastify tasks); the CodeRankEmbed side is not.
The quality claim above rests on the model author's published benchmarks, not on Vera's corpus.

citron07r:
## What

Makes CodeRankEmbed the default local embedding model (was jina-embeddings-v5-text-nano-retrieval), and pins its CoreML execution to CPU since the CoreML EP cannot run it at all.

## Why

Vera is a code-search tool but defaulted to a general-text embedding model. Published numbers from the model card ([huggingface.co/nomic-ai/CodeRankEmbed](https://huggingface.co/nomic-ai/CodeRankEmbed)):

| Model | CodeSearchNet MRR | CoIR NDCG@10 | Params |
|---|---|---|---|
| CodeRankEmbed | 77.9 | 60.1 | 137M |
| jina-embeddings-v2-base-code | 67.2 | 58.4 | 239M |
| Arctic-Embed-M-Long (CodeRankEmbed's own general-purpose base) | 53.4 | 43.0 | — |

The model actually shipped as Vera's current default publishes no code-retrieval number at all. CodeRankEmbed is also smaller.

**No A/B on Vera's own eval corpus was completed.** The run aborted on the CoreML bug described below, so this quality claim rests entirely on the model author's published benchmarks above, not on Vera-specific measurement. That comparison would be a good follow-up.

**This changes the default and requires existing users to reindex** — the embedding dimension and model identity change, so a stored index built against jina is not compatible with CodeRankEmbed at query time.

## The CoreML blocker

CodeRankEmbed cannot run on the CoreML execution provider. Verified twice on Apple Silicon before this fix:

```
CoreMLExecutionProvider_..._1 node ... GetStaticOutputShape
CoreML static output shape ({2,1,1,80}) and inferred shape ({1,1,1,-1}) have an inconsistent static dimensions (2 vs. 1)
Error: indexing failed: embedding generation failed
```

Session creation succeeds, so this only surfaces once inference actually runs — the same class of bug as #41, where the reranker's CoreML EP accepted a fused subgraph and then died at inference. That was fixed with `reranker_execution_provider`, which maps CoreML → CPU for the reranker specifically.

This PR adds the embedding-side equivalent, `embedding_execution_provider` in `crates/vera-core/src/local_models/mod.rs`, wired into `build_session` in `crates/vera-core/src/embedding/local_provider.rs`. It pins **only CodeRankEmbed** to CPU under CoreML — jina and any custom override keep CoreML acceleration.

**Cost:** CoreML accelerates embeddings only (~39ms), so pinning CodeRankEmbed to CPU under CoreML costs indexing throughput on Apple Silicon relative to a model CoreML could run. The alternative is a hard indexing failure, so paying that cost is strictly better than the status quo.

## Fallout from flipping the default

`defaults_for_source`'s fallback for an arbitrary custom HuggingFace repo or directory source used to reuse `Self::default()` as its field template. With `default()` now CodeRankEmbed, that silently gave custom models CodeRankEmbed's CLS pooling and required query prefix instead of jina's mean-pooling/no-prefix shape. Decoupled it to use `Self::jina()` explicitly as the generic template, restoring prior behavior for custom repos. Caught by the existing `gpu_adjustment_only_changes_the_default_jina_model` test.

## Verification

```
$ cargo build --release -p vera-cli   # clean
$ cargo fmt --check                    # clean
$ cargo clippy -p vera-core --lib      # 5 warnings, all pre-existing (unrelated to this change)
$ cargo test --workspace               # 917 passed, 0 failed
```

End-to-end against the built binary, default backend, no explicit `--onnx-*` flag (`VERA_NO_UPDATE_CHECK=1` set per contributor docs):

````
$ printf 'export function addNumbers(a: number, b: number): number { return a + b; }\n' > a.ts
$ vera index .
Indexing complete!

  Files parsed:        1
  Chunks created:      1
  Embeddings generated: 1
  Elapsed time:        0.37s

$ vera search "add two numbers"
```a.ts:1-1 function:addNumbers
export function addNumbers(a: number, b: number): number { return a + b; }
```
````

Indexing on the default backend, which previously died with the CoreML shape-mismatch error above, now succeeds, and a natural-language query returns the correct semantic match.

## Refs

References #70, #71.

---
## Summary by cubic
Sets `CodeRankEmbed` as the default local embedding model and pins it to CPU under CoreML to avoid an inference-time shape-mismatch crash. The old default was `jina-embeddings-v5-text-nano-retrieval`; existing indices are incompatible and must be rebuilt.

- CoreML behavior: only `CodeRankEmbed` is CPU-pinned via `embedding_execution_provider`; other embedding models keep CoreML acceleration.
- Provider resolution: single-sourced in `embedding_execution_provider` + `resolve_provider_and_config`; `build_session` trusts the resolved provider. Probes share this resolution. `vera doctor` now preflights the effective embedding/reranker providers (not just the requested backend), loads the correct ORT library, inspects assets for the effective EP, and runs dependency checks once per distinct non-CPU provider; the repair hint stays on the requested EP. It also respects the single-library load semantics of `ensure_ort_runtime`.
- Diagnostics: `vera doctor` consumes the already-resolved embedding config, removes the unreachable “unknown placement” branch, and for CoreML reports whether embeddings are CPU-pinned without claiming confirmed GPU execution.
- Custom sources: `from_huggingface_repo`/`from_directory` route through `defaults_for_source` using a Jina template so arbitrary repos/directories keep mean pooling and no query prefix (do not inherit `CodeRankEmbed`’s CLS pooling/query prefix).

**Migration**
- Rebuild any index created with the Jina default before upgrading.

<sup>Written for commit 6475510281c4b9e95fae6c69da8c1acc79cecd6a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/72?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * CodeRankEmbed is now the default local embedding model.
  * Source-specific configurations retain appropriate defaults, including Jina where applicable.
  * Custom repositories and local model directories now use their intended embedding behavior.

* **Bug Fixes**
  * Improved provider selection for CoreML compatibility, including CPU fallback where required.
  * Enhanced embedding provider initialization and inference reliability.
  * Diagnostics now separately report embedding and reranking provider readiness.
  * Updated guidance clearly identifies CPU usage for embedding and reranking.
  * Preserved correct configuration behavior for custom embedding models.

Fixes #70
Fixes #71

### DIFF
--- a/crates/vera-cli/src/commands/doctor.rs
+++ b/crates/vera-cli/src/commands/doctor.rs
@@ -21,6 +21,46 @@ struct DoctorCheck {
     detail: String,
 }
 
+/// Execution providers the local ONNX sessions actually run on.
+///
+/// A single requested backend does not decide this: the embedding session and
+/// the reranker session each resolve it independently, and either can land on
+/// CPU while the request was a GPU provider. Preflighting the request instead
+/// of the resolution makes doctor report failures for a configuration
+/// production runs fine, and vice versa.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct EffectiveProviders {
+    embedding: vera_core::config::OnnxExecutionProvider,
+    reranker: vera_core::config::OnnxExecutionProvider,
+}
+
+impl EffectiveProviders {
+    fn resolve(
+        ep: vera_core::config::OnnxExecutionProvider,
+        embedding_model: &vera_core::local_models::LocalEmbeddingModelConfig,
+    ) -> Self {
+        Self {
+            embedding: vera_core::local_models::embedding_execution_provider(ep, embedding_model),
+            reranker: vera_core::local_models::reranker_execution_provider(ep),
+        }
+    }
+
+    /// Distinct non-CPU providers needing a shared-library dependency check,
+    /// in embedding-then-reranker order. Empty when both sessions run on CPU,
+    /// where provider-specific checks are not needed.
+    fn dependency_check_providers(self) -> Vec<vera_core::config::OnnxExecutionProvider> {
+        let mut providers = Vec::new();
+        for candidate in [self.embedding, self.reranker] {
+            if candidate != vera_core::config::OnnxExecutionProvider::Cpu
+                && !providers.contains(&candidate)
+            {
+                providers.push(candidate);
+            }
+        }
+        providers
+    }
+}
+
 #[derive(Debug, Clone, Serialize)]
 struct DoctorReport {
     version: String,
@@ -74,7 +114,13 @@ pub fn run(json_output: bool, probe: bool) -> anyhow::Result<()> {
                 status: CheckStatus::Ok,
                 detail: embedding_model.display_name(),
             });
-            let runtime_path = vera_core::local_models::ort_library_path_for_ep(ep)?;
+            // The embedding and reranker sessions each resolve the requested
+            // backend to their own provider, so preflight the resolved pair
+            // rather than the request: otherwise doctor checks a library
+            // production never loads and skips every probe when it is absent.
+            let providers = EffectiveProviders::resolve(ep, &embedding_model);
+            let runtime_path =
+                vera_core::local_models::ort_library_path_for_ep(providers.embedding)?;
             let runtime_check = vera_core::local_models::ensure_ort_runtime(Some(&runtime_path));
             let runtime_detail = match &runtime_check {
                 Ok(()) => runtime_path.display().to_string(),
@@ -90,16 +136,20 @@ pub fn run(json_output: bool, probe: bool) -> anyhow::Result<()> {
                 detail: runtime_detail,
             });
 
-            let model_assets =
-                vera_core::local_models::inspect_local_model_files_for_ep(ep, &embedding_model)?;
+            let model_assets = vera_core::local_models::inspect_local_model_files_for_ep(
+                providers.embedding,
+                &embedding_model,
+            )?;
+            // The hint stays on the requested backend: `vera repair
+            // --onnx-jina-<ep>` is what installs that backend's files.
             let repair_hint = format!("run `vera repair --onnx-jina-{ep}`");
             checks.push(local_model_assets_check(&model_assets, &repair_hint));
             if probe {
                 checks.extend(probe_local_backend(
                     ep,
+                    providers,
                     &runtime_path,
                     &model_assets,
-                    &embedding_model,
                 )?);
             }
         }
@@ -302,12 +352,22 @@ fn probe_potion_backend(
 
 fn probe_local_backend(
     ep: vera_core::config::OnnxExecutionProvider,
+    providers: EffectiveProviders,
     runtime_path: &std::path::Path,
     model_assets: &[vera_core::local_models::LocalModelAssetStatus],
-    embedding_model: &vera_core::local_models::LocalEmbeddingModelConfig,
 ) -> anyhow::Result<Vec<DoctorCheck>> {
+    let embedding_ep = providers.embedding;
+    let reranker_ep = providers.reranker;
     let mut checks = Vec::new();
 
+    // `ensure_ort_runtime` is a process-wide `OnceLock`: only the FIRST path
+    // handed to it is dlopened, and every later call replays that first result
+    // whatever path it is given. `runtime_path` is the embedding provider's
+    // library, so when the two components resolve to different providers the
+    // reranker's library can only be existence-checked (which is what
+    // `dependency_probe_status` does), never load-checked. Calling
+    // `ensure_ort_runtime` a second time would report the first library's
+    // outcome under the second library's name.
     let ort_stage = result_check(
         "probe-ort-library",
         runtime_path.display().to_string(),
@@ -316,40 +376,108 @@ fn probe_local_backend(
     let ort_ok = matches!(ort_stage.status, CheckStatus::Ok);
     checks.push(ort_stage);
 
+    let mut embedding_provider_ok = false;
+    let mut reranker_provider_ok = false;
     let provider_stage = if ort_ok {
-        result_check(
-            "probe-provider-registration",
-            format!("registered {}", ep),
+        let embedding_result =
             vera_core::embedding::local_provider::LocalEmbeddingProvider::probe_provider_registration(
-                ep,
-            ),
-        )
+                embedding_ep,
+            );
+        embedding_provider_ok = embedding_result.is_ok();
+        if reranker_ep == embedding_ep {
+            reranker_provider_ok = embedding_provider_ok;
+            result_check(
+                "probe-provider-registration",
+                format!("registered {embedding_ep} for embedding and reranking"),
+                embedding_result,
+            )
+        } else {
+            let reranker_result =
+                vera_core::embedding::local_provider::LocalEmbeddingProvider::probe_provider_registration(
+                    reranker_ep,
+                );
+            reranker_provider_ok = reranker_result.is_ok();
+            let failures = [
+                embedding_result
+                    .err()
+                    .map(|err| format!("embedding {embedding_ep}: {}", one_line_error(&err))),
+                reranker_result
+                    .err()
+                    .map(|err| format!("reranker {reranker_ep}: {}", one_line_error(&err))),
+            ]
+            .into_iter()
+            .flatten()
+            .collect::<Vec<_>>();
+            if failures.is_empty() {
+                DoctorCheck {
+                    name: "probe-provider-registration",
+                    status: CheckStatus::Ok,
+                    detail: format!(
+                        "registered {embedding_ep} for embedding and {reranker_ep} for reranking"
+                    ),
+                }
+            } else {
+                DoctorCheck {
+                    name: "probe-provider-registration",
+                    status: CheckStatus::Fail,
+                    detail: failures.join("; "),
+                }
+            }
+        }
     } else {
         skipped_check(
             "probe-provider-registration",
             "skipped because ONNX Runtime could not be initialized",
         )
     };
-    let provider_ok = matches!(provider_stage.status, CheckStatus::Ok);
     checks.push(provider_stage);
 
-    let dependencies_stage = if ep == vera_core::config::OnnxExecutionProvider::Cpu {
+    let dependency_providers = providers.dependency_check_providers();
+    let mut failed_dependency_providers = Vec::new();
+    let dependencies_stage = if dependency_providers.is_empty() {
         skipped_check(
             "probe-dependencies",
-            "skipped for the CPU backend because provider-specific shared-library checks are not needed",
+            "skipped because the embedding and reranker sessions both run on CPU, where provider-specific shared-library checks are not needed",
         )
     } else if ort_ok {
-        dependency_probe_check(ep, runtime_path)
+        let mut status = CheckStatus::Ok;
+        let mut details = Vec::new();
+        for provider in &dependency_providers {
+            let library_path = if *provider == embedding_ep {
+                runtime_path.to_path_buf()
+            } else {
+                vera_core::local_models::ort_library_path_for_ep(*provider)?
+            };
+            let (provider_status, detail) = dependency_probe_status(*provider, &library_path);
+            if matches!(provider_status, CheckStatus::Fail) {
+                failed_dependency_providers.push(*provider);
+            }
+            status = worse_status(status, provider_status);
+            details.push(if dependency_providers.len() > 1 {
+                format!("{provider}: {detail}")
+            } else {
+                detail
+            });
+        }
+        DoctorCheck {
+            name: "probe-dependencies",
+            status,
+            detail: details.join("; "),
+        }
     } else {
         skipped_check(
             "probe-dependencies",
             "skipped because ONNX Runtime could not be initialized",
         )
     };
-    let dependencies_ok = !matches!(dependencies_stage.status, CheckStatus::Fail);
+    let embedding_dependencies_ok = !failed_dependency_providers.contains(&embedding_ep);
+    let reranker_dependencies_ok = !failed_dependency_providers.contains(&reranker_ep);
     checks.push(dependencies_stage);
 
-    let embedding_session_stage = if ort_ok && provider_ok && dependencies_ok {
+    let embedding_ok = ort_ok && embedding_provider_ok && embedding_dependencies_ok;
+    let reranker_ok = ort_ok && reranker_provider_ok && reranker_dependencies_ok;
+
+    let embedding_session_stage = if embedding_ok {
         let missing = missing_assets(model_assets, &["embedding-onnx", "embedding-onnx-data"]);
         if missing.is_empty() {
             result_check(
@@ -373,7 +501,7 @@ fn probe_local_backend(
     };
     checks.push(embedding_session_stage);
 
-    let reranker_session_stage = if ort_ok && provider_ok && dependencies_ok {
+    let reranker_session_stage = if reranker_ok {
         let missing = missing_assets(model_assets, &["reranker-onnx"]);
         if missing.is_empty() {
             result_check(
@@ -397,7 +525,7 @@ fn probe_local_backend(
     };
     checks.push(reranker_session_stage);
 
-    let tiny_inference_stage = if ort_ok && provider_ok && dependencies_ok {
+    let tiny_inference_stage = if embedding_ok && reranker_ok {
         let missing = missing_assets(
             model_assets,
             &[
@@ -436,7 +564,7 @@ fn probe_local_backend(
     let tiny_inference_ok = matches!(tiny_inference_stage.status, CheckStatus::Ok);
     checks.push(tiny_inference_stage);
 
-    if ep != vera_core::config::OnnxExecutionProvider::Cpu {
+    if !dependency_providers.is_empty() {
         checks.push(if tiny_inference_ok {
             DoctorCheck {
                 name: "probe-provider-confirmation",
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by citron07r** [comment]
Thanks — the `from_directory` finding was correct, and it was the more serious of the two ways this could go wrong. Fixed in 530e12e.

**`from_directory` inheriting CodeRankEmbed (valid, fixed).** Both bots flagged this independently and both were right. `defaults_for_source` was added precisely so custom sources keep jina's mean-pooling/no-prefix shape, and its doc comment already claimed to cover directories — but `from_directory` still spread `..Self::default()`, and the `VERA_LOCAL_EMBEDDING_DIR` branch returns from `from_directory` directly, so that arm was unreachable. Both constructors now go through a shared `from_source`. Added `custom_sources_do_not_inherit_coderank_pooling_or_query_prefix`, covering the repo and directory cases; confirmed it fails with the old `..Self::default()` restored (`left: Cls, right: Mean`) and passes with the fix.

**Scaler/provider derived from the requested EP, not the effective one (valid, fixed).** This was the better of the two `local_provider.rs` findings. `build_session` remapped CoreML to CPU internally while `new_with_ep_and_mem_limit` kept branching on the requested EP, so a CodeRankEmbed session that actually ran on CPU still got the GPU adaptive batch path, GPU provider dependencies, and — worst of the three — an `ep != Cpu` "fallback" that would have retried the CPU session it had just built. The effective EP is now resolved once, up front.

While fixing that I found a related gap neither review caught: under CoreML, `vera doctor` asserted "only the embedding model is GPU-accelerated", which this PR makes false for the new default. It now reports what the configured model actually does.

**Pinning to CPU even when the ONNX file is overridden (valid, deliberately not changed).** Correct that the pin keys on the source repo, so overriding `VERA_LOCAL_EMBEDDING_ONNX_FILE` while keeping the CodeRankEmbed repo keeps the CPU pin. I'm leaving it, because the failure modes are asymmetric: an unnecessary pin costs throughput, whereas guessing that a replacement graph runs on CoreML risks a hard failure at inference deep inside indexing — which is exactly how #41 and #71 presented. A plausible override here is `onnx/model_fp16.onnx` for the same model, which is *more* likely to fail on CoreML, not less. Detecting "unmodified preset" is easy; knowing the replacement is CoreML-safe is not, and only the second would justify dropping the pin.

845 tests pass, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings. Re-verified end-to-end on the default backend: `vera index` succeeds and `vera search "add two numbers"` returns the expected match.

**Comment by citron07r** [comment]
Ran a second review pass locally (the hosted one was rate-limited). Two more findings, both valid, both fixed in 337dec4.

**Probe paths resolved the provider differently from production (fixed).** This was the better catch, and it's the same bug I'd just fixed one layer up. `probe_session`/`probe_inference` called `adjust_for_gpu(requested_ep)` and only then let `build_session` resolve to CPU, so for a CPU-pinned model `vera doctor` would probe a GPU-adjusted configuration that production never loads — a probe reporting on something other than what runs. Both now share `resolve_provider_and_config` with the constructor, so there is one place that answers "which provider, and which config for it".

Worth being precise about severity: the divergence is currently latent rather than active, because `adjust_for_gpu` only rewrites the model for the default jina model (`gpu_adjustment_only_changes_the_default_jina_model` pins that), and CodeRankEmbed — the only pinned model — is untouched by it. So no user hits this today. It becomes live the moment any pinned model has a GPU export, which is exactly the kind of thing that gets added later and breaks quietly. Fixed on that basis.

**Doctor claimed confirmed GPU execution (fixed).** Correct, and it contradicted the check immediately above it. Selecting the CoreML provider is not evidence ONNX Runtime placed the graph there — it can still assign nodes to CPU, which is the whole reason `probe-provider-confirmation` exists and says active GPU execution cannot be confirmed. The warning now says the embedding model is *configured* for CoreML and that GPU execution cannot be confirmed here, rather than asserting acceleration.

845 tests pass, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings. Re-ran the end-to-end check on the default backend after the change: `vera index` succeeds and `vera search "add two numbers"` returns the expected match.

**Comment by citron07r** [comment]
Third local pass, three more findings, all valid and all in code I'd added in the previous two rounds. Fixed in d9facc7.

**Two provider-resolution paths (fixed).** Right, and it was my own doing: I added `resolve_provider_and_config` last round but left `build_session` calling `embedding_execution_provider` independently. No behavioural bug today — the mapping is idempotent, so resolving twice gives the same answer — but it is the same contract written down twice, and the whole point of the previous fix was that these had already drifted once. `build_session` now trusts its caller. All three call sites resolve first, so this is verifiable rather than aspirational, and it drops the now-dead `config` parameter and a per-load `LocalEmbeddingModelConfig` clone with it.

**`is_ok_and` collapsed two states in the doctor message (fixed).** Good catch. It returned false both for a CPU-pinned model and for a config that could not be read, so an unreadable config would have been reported as "configured for CoreML" — a confident claim about something we failed to determine, which is the exact failure mode the previous round was fixing. Now three explicit states, with the unreadable case saying placement could not be determined.

**Tests for the contract (added).** Two of them, deliberately asset-free so they run when ONNX Runtime is absent — the neighbouring `test_local_embedding_provider` early-returns and passes without asserting anything, which is worth knowing when reading a green run here.

One note in the interest of not overstating: the third test I first wrote was worthless. It called `resolve_provider_and_config` twice and asserted the results matched, which is true by construction — the same tautology CodeRabbit correctly flagged in #73's ordering test. I replaced it with an idempotency assertion, which is the property that actually has to hold now that `build_session` no longer re-resolves.

848 tests pass, `cargo fmt --check` clean, clippy back to exactly the 5 pre-existing warnings (the parameter removal briefly added two, both now gone rather than silenced). Re-ran the end-to-end check on the default backend: index and search both behave.

**Comment by citron07r** [comment]
Two findings on the same block, and the second one shows the first fix was wrong. Both resolved in e6499db.

**`is_ok_and` hid configuration errors (already fixed in d9facc7).** Correct, and this was fixed in the previous round: it became a three-way `match` over `Some(true)` / `Some(false)` / `None`, so an unreadable config reported that placement could not be determined rather than claiming CoreML.

**That third branch was unreachable (fixed properly now).** This is the better catch and it retires the fix above. `run` resolves the embedding config with `from_env()?` at the top of the local-backend arm and propagates the failure, so `probe_local_backend` is never reached when the config cannot be read. My `None` arm could not fire — I had traded a boolean that hid errors for a match arm that was dead code, which is not an improvement.

`probe_local_backend` now takes the already-validated config, which is the shape suggested. That removes the dead branch *and* the second `from_env()` read, which was the deeper problem: two reads of the same source in one command can disagree, and the check would have been reporting on a different config than the one `run` validated and displayed as `local-embedding-model`.

Verified end-to-end on a CoreML backend with the jina model, since the two arms are easy to get backwards:

```
warn  probe-reranker-coreml-cpu  the reranker runs on CPU under CoreML (...). The embedding
      model is configured for CoreML, but ONNX Runtime can still place nodes on CPU, so
      active GPU execution cannot be confirmed here.
```

That is the correct arm — jina is not CPU-pinned, so it reports configured-for-CoreML with execution unconfirmed rather than claiming acceleration.

848 tests pass (756 `vera-core`, 92 `vera-cli`), `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

**Comment by citron07r** [comment]
Valid, and the structural diagnosis is exactly right. Fixed in 6475510. Two corrections to the specifics, both of which cut in favour of fixing it.

**The named symptom is latent; a different instance of the same defect is live.** CoreML *registration* failing while production succeeds is not reachable in practice: on macOS arm64 both Cpu and CoreMl pull the identical `onnxruntime-osx-arm64` archive with the CoreML EP compiled in, and `probe_provider_registration` only builds a `SessionBuilder` and registers the EP without committing a model. The dependency check is similarly inert, since `macos_dependency_exists` accepts anything under `/usr/lib` or `/System/Library`.

What *is* reachable is the ORT library path. `preferred_ort_library_path_for_ep_in_home` gives `~/.vera/lib/libonnxruntime.dylib` for Cpu and `~/.vera/lib/<ep>/libonnxruntime.dylib` otherwise. If the CoreML copy is absent while the CPU one is present (backend switched without re-running setup, or a failed download), doctor failed the ORT check and skipped every probe, while production resolved CodeRankEmbed to Cpu and loaded the CPU library fine.

The inverse is worse and also reachable: after `vera setup --onnx-jina-coreml` with CodeRankEmbed, only the CoreML copy is installed, so doctor reported `onnx-runtime ok` for a library production will never load, and the first real `vera index` silently downloaded a second archive. Offline, that is a hard failure doctor had rated green.

**The asset-list half of the claim does not hold.** `inspect_local_model_files_for_ep` does call `adjust_for_gpu(ep)`, but that only swaps the quantized export for fp16 when the repo is the jina one, and jina is never CPU-pinned. CodeRankEmbed, the only model that resolves CoreMl to Cpu, is unaffected at every provider, so the embedding asset list is identical for requested and effective ep in every configuration that exists. The one entry that did diverge is the ORT library path, which is the same bug as above rather than a separate fp16 concern.

**What changed.** An `EffectiveProviders { embedding, reranker }` resolved once in `run` and threaded down, so the command cannot disagree with itself:

- the ORT library path and asset inspection follow the effective embedding provider (the repair hint stays on the requested one, since `vera repair --onnx-jina-coreml` is what installs that backend's files)
- `probe-provider-registration` probes the embedding provider, plus the reranker's when they differ, and names them
- `probe-dependencies` skips only when both are Cpu, otherwise runs once per distinct non-Cpu provider
- the single `ort_ok && provider_ok && dependencies_ok` gate is now per component, so a reranker-side failure no longer suppresses the embedding probe
- `probe-provider-confirmation` fires when either effective provider is non-Cpu

One constraint worth recording, now commented at the call site: `ensure_ort_runtime` is a process-wide `OnceLock`, so only the first path passed to it is actually loaded. When the two components resolve to different providers the second library can only be existence-checked, not load-checked. Calling it twice and reporting the second result would be reporting something untrue.

**You were right that the reranker needs the same treatment, and it is a deeper problem than the doctor.** `LocalReranker::new_with_ep` acquires, initializes and dependency-checks the library for the *requested* provider and only resolves inside `build_session`. Combined with the `OnceLock`, it can validate the CoreML dylib and then build against whichever library the embedding path already loaded, so a green check is not evidence. That changes download and network behaviour, so I filed it as #89 rather than folding it into a PR about model defaults.

Four pure tests cover the resolution and gating decision, chosen because anything touching `ensure_ort_runtime` skips silently without ORT present: CodeRankEmbed on CoreML (both to Cpu, no dependency check), jina on CoreML (embedding stays CoreMl, reranker Cpu, check needed), Cpu with any model, and a non-CoreML GPU backend.

Verified end to end on this Apple Silicon machine with the CoreML backend, where every probe line is still `ok` and registration now names the effective providers:

```
ok  probe-provider-registration  registered coreml for embedding and cpu for reranking
ok  probe-dependencies           found no unresolved ONNX Runtime dependencies
ok  probe-embedding-session      embedding session created
ok  probe-reranker-session       reranker session created
ok  probe-tiny-inference         embedding and reranker returned finite outputs
```

Being straight about coverage: the reachable missing-dylib case is covered by unit test, not reproduced live, because doing so would mean deleting a library from the real `~/.vera` install.

756 `vera-core` and 96 `vera-cli` tests pass, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

lemon07r:
Closing without merge: superseded by benchmark evidence.

The 2026-08-21 default-model screening (docs/benchmarks.md, "Default-Model Screening"; subset 320 + full 1,251 Semble tasks, no rerank, CUDA, all models pinned by commit SHA) found no quality premium for CodeRankEmbed over the field at Vera's 512-token truncation:

- Full suite: jina v5 nano 0.7452 nDCG, CodeRankEmbed 0.7401, potion-code-16M-v2 0.7388 — all within ~0.006.
- The published +10.7 CSN advantage (#70) did not transfer to our task set.

With quality at parity, the decision turned on operational cost and license: potion-code-16M-v2 (MIT, 16M static) indexes the 63-repo suite in 443s vs CodeRankEmbed's 3308s and halves query p50 (16.8ms vs 27.0ms). We're making potion-code-16M-v2 the default local embedding model instead (native model2vec backend, one reindex wave).

Thanks for pushing the licensing issue — the CC-BY-NC default embedding concern (#66) gets resolved by the potion-v2 switch too. The CoreML pin work here (embedding_execution_provider / EffectiveProviders) was independently good and may return if CodeRankEmbed stays an opt-in preset; #71 stays open for that.

PR #73, opened 2026-08-19
perf(indexing): parallelize update path and batch metadata writes/reads (#69)

### LINKED ISSUE #69: Incremental update path is single-threaded while full index uses rayon, plus N+1 commits and per-candidate chunk fetches
Audit of parallelism and threading across the indexing path, the search path and the ONNX session
setup. Every finding below is grounded in a file:line reference. Nothing here is measured unless it
says so, and several things that looked like problems turned out to be already correct — those are
recorded too, so nobody re-investigates them.

## The main finding: the update path is single-threaded, the full-index path is not

`pipeline.rs:380` parses a full index in parallel:

```rust
let results: Vec<ParsedFileResult> = discovery.files.par_iter().map(|file| { ... })
```

`update.rs:400` parses an incremental update serially:

```rust
for (rel_path, content, _hash) in &files_to_index {
```

That `par_iter` at `pipeline.rs:380` is the **only** one in all of `indexing/` and `discovery/`.
Tree-sitter parsing is CPU-bound, so `vera update` leaves every core but one idle while
`vera index` does not.

| # | location | problem |
|---|---|---|
| 1 | `update.rs:288-298`, `freshness.rs:113-136` | every current file is read sequentially before anything is classified, including unchanged ones, so `update` and staleness checks are I/O-bound on total repo size rather than on the size of the change |
| 2 | `update.rs:399-456` | serial tree-sitter parse of modified/added files; the full-index path already shows the rayon boundary to copy |
| 3 | `discovery/mod.rs:219-336` | single-threaded `walker.build()` plus a per-entry `metadata` stat and 8KB binary sniff; the `ignore` crate's `WalkBuilder::build_parallel()` is unused |
| 4 | `pipeline.rs:598-620` | per-file `set_file_hash`, `insert_references` and `insert_type_relations` each commit their own transaction (`metadata.rs:577-596`, `604-630`), so N files means N+1 commits |
| 5 | `retrieval/vector.rs:80-94` | `metadata_store.get_chunk` per candidate in a loop — N+1 round-trips; `get_chunk` (`metadata.rs:271-289`) is single-row and there is no batch variant |

## Things that are already correct

Recording these so they are not re-audited:

- **BM25 and vector search genuinely overlap.** `hybrid.rs:109` spawns BM25 via `spawn_blocking` and does not await it until line 174, with the vector arm running inline in between on a separate `MetadataStore` connection. The doc comment is accurate.
- **The reranker is batched**, not per-pair: `local_reranker.rs:13,196` uses `MAX_RERANK_BATCH_SIZE = 8` with one padded `Array2` and one `session.run()` per batch.
- **ONNX intra-op threads are set** (`local_reranker.rs:255-260`, `local_provider.rs:821-825`), sessions are built once and reused behind `Arc<Mutex<Session>>`, and `GraphOptimizationLevel::Level3` is set at both construction sites.
- `insert_chunks` (`metadata.rs:237-269`) and `VectorStore::insert_batch` (`vector.rs:136-190`) batch correctly in a single transaction with `prepare_cached`.

## Smaller opportunities

- **`local_provider.rs:962-966`** — the CoreML EP is registered as `CoreMLExecutionProvider::default()` with no `.with_compute_units()`. Explicitly requesting the Neural Engine may move matmul-heavy embedding work off CPU/GPU on Apple Silicon. Unmeasured.
- **`local_provider.rs:34-124` (`AdaptiveBatchScaler`)** — growth is geometric after a success but `min_failed_batch` is sticky, so a single early failure on a noisy first request can cap a sequence-length bucket for the life of the process until a later success in the same bucket clears it.
- **`local_provider.rs:821-825`** — GPU EPs get `available.min(4)` intra-op threads. The `4` has no derivation in a comment.
- **`MAX_RERANK_BATCH_SIZE = 8`** against a default 50 rerank candidates means roughly 7 sequential forward passes per search. Raising it trades padding cost against pass count; worth measuring given reranking is ~98% of search latency.

## Note for #67

Raising `EMBEDDING_MAX_LENGTH` is not a standalone constant bump. `AdaptiveBatchScaler::bucket_for`
(`local_provider.rs:103-106`) derives its bucket width from `max_length / 8`, so the batch scaler's
bucketing and cold-start estimate need re-tuning in the same change.

## Caveat

These are static-analysis findings. None of the wins is quantified, and the ranking is by expected
impact rather than measurement. The rayon asymmetry between the two parse paths is the one I would
treat as clear-cut, since the codebase already contains the correct pattern.

citron07r:
## Summary

Implements all four performance fixes from #69:

1. **Parallelize serial file parsing in the update path** (`indexing/update.rs`) — tree-sitter parsing of modified/added files was a serial `for` loop; now uses rayon `par_iter`, mirroring the shape already used by the full-index path in `pipeline.rs`.
2. **Parallelize sequential file reads for freshness/update classification** (`indexing/update.rs`, `indexing/freshness.rs`) — file reads (and, in freshness.rs, the content hashing) are now done with rayon `par_iter` instead of a serial loop over every current file.
3. **Batch per-file transactions into one transaction** (`storage/metadata.rs`, `indexing/pipeline.rs`, `indexing/update.rs`) — added `MetadataStore::set_file_hashes_batch` and `MetadataStore::insert_parse_artifacts_batch`, each wrapping the previous per-file `set_file_hash` / `insert_references` / `insert_type_relations` calls in a single transaction. Used from both the full-index and incremental-update paths. The existing single-file methods are left in place (still used by tests and elsewhere).
4. **Batch chunk fetch to eliminate N+1 query** (`retrieval/vector.rs`, `storage/metadata.rs`) — added `MetadataStore::get_chunks_by_ids` (single `SELECT ... WHERE id IN (...)`), used in `search_vector_with_stores` instead of one `get_chunk` call per candidate.

Nothing was skipped — all four items from the issue landed.

## Correctness

These are correctness-preserving refactors; no ranking/retrieval-result change, so no Semble benchmark evidence is included (relevance ordering is unchanged).

For fix 4 specifically: the same chunks are returned in the same order as before. SQLite does not guarantee row order for `IN (...)`, so the batch result is collected into a lookup map and re-projected back into the original vector-distance order (the order candidates arrived in from the vector store), preserving the existing early-break-at-`limit` behavior. Only the number of SQL round-trips changes (N queries → 1 batch query). A regression test (`batch_chunk_fetch_preserves_vector_distance_order` in `retrieval/vector.rs`) inserts chunks out of request order and asserts results come back in the requested order, not storage/insertion order.

For fixes 1 and 2, metadata-store writes that happened inside the original per-file loops are collected from the parallel parsing results and applied sequentially afterward — `MetadataStore` wraps a single non-`Sync` SQLite `Connection`, so it can't be called concurrently from multiple rayon worker threads.

## Verification

Run from a git worktree on `perf/parallel-update-path`, cut from `origin/master`.

- `cargo build -p vera-core --lib` — clean build after each fix.
- `cargo fmt --check` — clean, zero diffs.
- `cargo clippy -p vera-core --lib` — exactly the 5 pre-existing warnings (`pip_package_for_ep`, `CUDA_RUNTIME_LIBRARY_PREFIXES`, `parse_cuda_major_from_runtime_library_entry`, one unused import, one unused variable, all in `local_models/`), no new ones introduced.
- `cargo test --workspace` — **921 passed, 0 failed, 0 ignored** across the workspace (`vera` 92, `vera_core` 756, `vera_eval` 35, `vera_mcp` 37, `vera_serve` 1; doc-tests 0/0/0).

## Test plan

- [x] `cargo build -p vera-core --lib`
- [x] `cargo fmt --check`
- [x] `cargo clippy -p vera-core --lib` (no new warnings vs. baseline)
- [x] `cargo test --workspace` (921/921 passing)
- [x] New regression test added for batch chunk fetch order preservation

---
## Summary by cubic
Parallelizes incremental update and freshness scans and batches metadata I/O, while preserving write-after-embedding atomicity; retrieval now batch-fetches chunk metadata and preserves ranking. Old behavior: serial per-file parse/commits and an N+1 chunk fetch; new behavior: `rayon`-parallel reads/parsing and single-transaction batch writes/reads with identical results, order, and cancellation semantics.

- Update path: file reads and hashing run under `rayon`; parsing/chunking uses `par_iter` and preserves input order; `chunk_file_for_update` returns a per-file parse error for collection; unreadable files keep existing index data; writes move to `insert_parse_artifacts_batch` and `set_file_hashes_batch` (borrowed variants avoid copies) and still occur only after embeddings succeed.
- Freshness: parallelizes file reads and content hashing; counts unreadable tracked files as modified; keeps `MetadataStore` lookups sequential due to a single SQLite connection.
- Retrieval: replaces per-candidate `get_chunk` with `get_chunks_by_ids`; re-projects to vector-distance order; regression test asserts each chunk remains paired with its own distance.
- Storage: adds `get_chunks_by_ids`, `set_file_hashes_batch`, and `insert_parse_artifacts_batch` (single-file methods delegate); early-returns on empty input; tests cover multi-row persistence, replace-on-repeat, partial-result behavior, and cross-file reference persistence.
- Migration: none.

<sup>Written for commit 03eaeed51b6af7ae8baa45b2d0ffc59a43551160. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/73?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Fixes #69

## Summary by CodeRabbit

* **Performance**
  * Faster indexing and incremental updates through parallel file processing.
  * More efficient persistence of file metadata and parsed results through batched operations.
  * Faster vector search by retrieving matching chunk details in batches.

* **Reliability**
  * Preserved search ranking and result ordering.
  * Unreadable files continue to be skipped with warnings.
  * Parsing errors are reported per file without interrupting other updates.

### DIFF
--- a/crates/vera-core/src/indexing/freshness.rs
+++ b/crates/vera-core/src/indexing/freshness.rs
@@ -5,6 +5,7 @@ use std::path::{Path, PathBuf};
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use anyhow::{Context, Result};
+use rayon::prelude::*;
 use tracing::warn;
 
 use crate::config::IndexingConfig;
@@ -125,25 +126,45 @@ fn count_modified_files(
     metadata_store: &MetadataStore,
     repo_root: &Path,
 ) -> Result<usize> {
-    let mut files_modified = 0usize;
-    for (rel_path, absolute_path) in current_files
+    let tracked_current: Vec<(&String, &PathBuf)> = current_files
         .iter()
         .filter(|(path, _)| tracked_files.contains(path.as_str()))
-    {
-        let content = match crate::discovery::read_source_lossy(absolute_path) {
-            Ok(content) => content,
-            Err(err) => {
-                warn!(
-                    file = %rel_path,
-                    error = %err,
-                    "failed to read file during freshness scan"
-                );
-                files_modified += 1;
-                continue;
-            }
+        .collect();
+
+    // Reading and hashing is I/O and CPU bound, so it runs under rayon. The
+    // metadata lookup that follows stays sequential: `MetadataStore` wraps a
+    // single SQLite connection and is not `Sync`, so it cannot be called from
+    // several rayon threads at once.
+    //
+    // `None` means the file could not be read. It is counted as modified
+    // rather than skipped, so an unreadable tracked file cannot make a stale
+    // index look current (#74).
+    let hashed: Vec<(&String, Option<String>)> = tracked_current
+        .par_iter()
+        .map(|(rel_path, absolute_path)| {
+            let content = match crate::discovery::read_source_lossy(absolute_path) {
+                Ok(content) => content,
+                Err(err) => {
+                    warn!(
+                        file = %rel_path,
+                        error = %err,
+                        "failed to read file during freshness scan"
+                    );
+                    return (*rel_path, None);
+                }
+            };
+            let language = detect_language_for_path(rel_path);
+            let current_hash = hash_for_indexing_source(&content, rel_path, language, repo_root);
+            (*rel_path, Some(current_hash))
+        })
+        .collect();
+
+    let mut files_modified = 0usize;
+    for (rel_path, current_hash) in hashed {
+        let Some(current_hash) = current_hash else {
+            files_modified += 1;
+            continue;
         };
-        let language = detect_language_for_path(rel_path);
-        let current_hash = hash_for_indexing_source(&content, rel_path, language, repo_root);
         let stored_hash = metadata_store
             .get_file_hash(rel_path)
             .with_context(|| format!("failed to read stored hash for {rel_path}"))?;
--- a/crates/vera-core/src/indexing/pipeline.rs
+++ b/crates/vera-core/src/indexing/pipeline.rs
@@ -610,29 +610,21 @@ fn store_index(
         .insert_chunks(chunks)
         .context("failed to insert chunk metadata")?;
 
-    // Store file content hashes for incremental indexing.
-    for (file_path, hash) in file_hashes {
-        metadata_store
-            .set_file_hash(file_path, hash)
-            .context("failed to store file hash")?;
-    }
+    // Store file content hashes for incremental indexing, batched into a
+    // single transaction instead of one commit per file.
+    metadata_store
+        .set_file_hashes_batch(file_hashes)
+        .context("failed to store file hashes")?;
 
     metadata_store
         .insert_file_states(metadata.file_states)
         .context("failed to store file index states")?;
 
-    // Store call-site references for call graph analysis.
-    for (file_path, refs) in file_refs {
-        metadata_store
-            .insert_references(file_path, refs)
-            .context("failed to store references")?;
-    }
-
-    for (file_path, relations) in file_type_relations {
-        metadata_store
-            .insert_type_relations(file_path, relations)
-            .context("failed to store type relations")?;
-    }
+    // Store call-site references and type relations for call graph analysis,
+    // batched into a single transaction instead of up to two commits per file.
+    metadata_store
+        .insert_parse_artifacts_batch(file_refs, file_type_relations)
+        .context("failed to store references and type relations")?;
 
     metadata_store
         .set_index_meta("model_name", metadata.model_name)
--- a/crates/vera-core/src/indexing/update.rs
+++ b/crates/vera-core/src/indexing/update.rs
@@ -19,6 +19,7 @@ use std::path::Path;
 use std::time::Instant;
 
 use anyhow::{Context, Result, bail};
+use rayon::prelude::*;
 use sha2::{Digest, Sha256};
 use tracing::{debug, info, warn};
 
@@ -29,6 +30,8 @@ use crate::embedding::{
     EmbeddingError, EmbeddingProvider, embed_chunks_concurrent_with_progress_and_cancellation,
 };
 use crate::parsing;
+use crate::parsing::references::RawReference;
+use crate::parsing::type_relations::RawTypeRelation;
 use crate::storage::bm25::Bm25Index;
 use crate::storage::metadata::{FileIndexState, FileIndexStatus, MetadataStore};
 use crate::storage::vector::VectorStore;
@@ -451,18 +454,27 @@ where
     let mut parse_errors = Vec::new();
 
     if !files_to_index.is_empty() {
-        // Parse and chunk new/modified files.
-        for (rel_path, content, hash, is_modified) in &files_to_index {
-            cancellation.check()?;
-            let language = detect_language_for_path(rel_path);
+        // Parse and chunk new/modified files. Tree-sitter parsing is CPU-bound,
+        // so this is parallelized with rayon (mirrors pipeline.rs's full-index
+        // path). Results are still staged in `PreparedFile` and written only
+        // after embedding succeeds. Parallelism here must not move any write
+        // earlier.
+        let parsed: Vec<(PreparedFile, Option<FileError>)> = files_to_index
+            .par_iter()
+            .map(|(rel_path, content, hash, is_modified)| {
+                cancellation.check()?;
+                let language = detect_language_for_path(rel_path);
 
-            // For RST, refs come from raw source; chunks from preprocessed.
-            // For all other languages, parse once for both.
-            let (chunks, refs, file_state) = if language == Language::Rst {
-                let refs = parsing::parse_and_extract_references(content, language);
-                let absolute_path = repo_root.join(rel_path);
-                let normalized_source =
-                    match parsing::sphinx::preprocess_rst(content, &absolute_path, &repo_root) {
+                // For RST, refs come from raw source; chunks from preprocessed.
+                // For all other languages, parse once for both.
+                let (chunks, refs, file_state, parse_error) = if language == Language::Rst {
+                    let refs = parsing::parse_and_extract_references(content, language);
+                    let absolute_path = repo_root.join(rel_path);
+                    let normalized_source = match parsing::sphinx::preprocess_rst(
+                        content,
+                        &absolute_path,
+                        &repo_root,
+                    ) {
                         Ok(preprocessed) => Some(preprocessed),
                         Err(err) => {
                             warn!(
@@ -473,38 +485,45 @@ where
                             None
                         }
                     };
-                let src = normalized_source.as_deref().unwrap_or(content);
-                chunk_file_for_update(
-                    src,
-                    rel_path,
-                    language,
-                    config,
-                    Some(refs),
-                    &mut parse_errors,
-                )
-            } else {
-                chunk_file_for_update(content, rel_path, language, config, None, &mut parse_errors)
-            };
+                    let src = normalized_source.as_deref().unwrap_or(content);
+                    chunk_file_for_update(src, rel_path, language, config, Some(refs))
+                } else {
+                    chunk_file_for_update(content, rel_path, language, config, None)
+                };
 
-            let type_relations = parsing::type_relations::extract_type_relations(&chunks);
+                let type_relations = parsing::type_relations::extract_type_relations(&chunks);
 
-            debug!(
-                file = %rel_path,
-                chunks = chunks.len(),
-                refs = refs.len(),
-                type_relations = type_relations.len(),
-                "parsed file"
-            );
+                debug!(
+                    file = %rel_path,
+                    chunks = chunks.len(),
+                    refs = refs.len(),
+                    type_relations = type_relations.len(),
+                    "parsed file"
+                );
 
-            prepared_files.push(PreparedFile {
-                path: rel_path.clone(),
-                hash: hash.clone(),
-                modified: *is_modified,
-                chunks,
-                references: refs,
-                type_relations,
-                state: file_state,
-            });
+                Ok((
+                    PreparedFile {
+                        path: rel_path.clone(),
+                        hash: hash.clone(),
+                        modified: *is_modified,
+                        chunks,
+                        references: refs,
+                        type_relations,
+                        state: file_state,
+                    },
+                    parse_error,
+                ))
+            })
+            .collect::<Result<_>>()?;
+        cancellation.check()?;
+
+        // `files_to_index` order is preserved by `collect`, so the sequential
+        // and parallel paths produce identical `prepared_files` ordering.
+        for (prepared, parse_error) in parsed {
+            prepared_files.push(prepared);
+            if let Some(parse_error) = parse_error {
+                parse_errors.push(parse_error);
+            }
         }
     }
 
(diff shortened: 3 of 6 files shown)


citron07r:
Five findings, three acted on and two declined with reasons. Fixes in 69d1b40.

**The ordering test was tautological (valid, and worse than described — fixed).** This is the one that mattered, and the criticism was right on every point. The test re-implemented the projection in its own body and compared a `HashMap` lookup against the key it looked up with, which holds by construction. Its end-to-end half only asserted that scores descend — and scores descend for *any* implementation that emits one result per candidate in candidate order, because `VectorStore::search` already returns rows ordered by distance. So the assertion could not fail on the bug it was named for.

It now calls `search_vector_with_stores` and asserts each chunk comes back paired with **its own** distance, plus `results.len()`, so dropped candidates fail loudly too.

Two things worth reporting. First, I added guards asserting the fixture's distance order differs from both insertion order and primary-key order — and the guard immediately failed: the original fixture's distance order was *identical* to its insertion order, so even a correct pairing assertion would have been trivially satisfied. The chunks are now ordered so the expected result is a genuine permutation of both.

Second, per this repo's verification standard I injected the regression before trusting the test. Zipping candidates against SQL order produces:

```
left:  [("src/a.rs", 0.433), ("src/b.rs", 0.378), ("src/c.rs", 0.353)]
right: [("src/c.rs", 0.433), ("src/a.rs", 0.378), ("src/b.rs", 0.353)]
```

Note the left column is still strictly descending — which is precisely why the old monotonicity check could never have caught this.

**Duplicated INSERT statements (valid, fixed).** `set_file_hash`, `insert_references` and `insert_type_relations` now delegate to the batch methods, following the `upsert_file_state`/`insert_file_states` precedent already in the file. One copy of each statement remains. Worth flagging for the maintainers: after this PR all three single-item methods have **no production callers** — only tests — so you may prefer to delete them outright rather than keep them as wrappers. I kept them because removing public API felt like your call, not something a perf PR should decide.

**Redundant `Arc` in the update path (valid, fixed).** Correct that it added no capability: rayon's closure needs `Sync`, and `VeraConfig`/`PathBuf` already are, so the `Arc` bought a `VeraConfig` clone per run and nothing else. Removed, along with the now-unused import. Note `pipeline.rs:366-367` does the same thing in `parse_discovered_files_parallel`; I left it alone rather than widen this PR into pre-existing code, but it is the same one-line cleanup if you want it.

**Unreadable files dropped from the freshness result (valid concern, but pre-existing — not changed here).** The underlying problem is real: if every tracked file fails to read, `detect_staleness` reports zero changes and a caller keeps a stale index. But this PR did not introduce it. On master the skip is a `continue` in the sequential loop; this PR moved the identical skip into the rayon closure as a `None` that the collection loop then drops. Same warning, same result, no error propagated either way. Fixing it means changing `detect_staleness`'s error contract, which does not belong in a perf PR — filed as #74 with three candidate shapes.

**Chunking the `IN (...)` list (declined — not reachable).** The bound holds with a wide margin. `ids` is one per element of `vector_results`, and `VectorStore::search` clamps to `MAX_KNN_K = 4096` *before* querying, so `ids.len() <= 4096` regardless of how large the candidate pool multipliers in `hybrid.rs` get. The workspace pins `rusqlite 0.39` with `bundled`, resolving `libsqlite3-sys 0.37` / SQLite 3.51.3, where `SQLITE_MAX_VARIABLE_NUMBER` defaults to **32766** and is only overridden via a build-time env var this repo does not set. That is 4096 against 32766 — 8x headroom. The 999 figure applies to SQLite before 3.32.0, which is not what ships here. Chunking would be dead code guarding an unreachable state.

848 tests pass, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

citron07r:
Force-pushed a rebase onto current master (#58 and #60 landed in the meantime). The shape of the change is materially different, so it is worth explaining rather than just noting the rebase.

**Why it changed.** #60 restructured exactly the function this PR parallelizes: parsed files are now staged in `PreparedFile` and every destructive write is deferred until embedding has succeeded, so a read or provider failure leaves the previous index intact. This PR, as originally written, cleared parse data up front and wrote references and type relations immediately after parsing — which would have reopened the window #60 had just closed.

That is not a hypothetical. I resolved the conflict the naive way first and ran your regression test against it:

```
update_embedding_failure_preserves_existing_parse_data ... FAILED
  left: 0
 right: 1
```

Zero references surviving where one should. So the textual resolution would have silently reverted #60 while looking like a clean merge.

**What I did instead.** The parallelism is now applied *inside* #60's structure rather than around it. Parsing runs under rayon and produces `PreparedFile` values; the writes stay at exactly the same point in the sequence and are only batched:

- Tree-sitter parsing is CPU-bound and runs in parallel. `collect` preserves `files_to_index` order, so `prepared_files` ends up ordered identically to the sequential path — worth stating explicitly since the deletion loop iterates it.
- `chunk_file_for_update` returns its parse error instead of pushing into a shared `&mut Vec<FileError>`. That shared borrow was the only thing preventing the loop from moving under rayon, and returning the error is what a parallel map needs anyway.
- The per-file `insert_references`/`insert_type_relations` loop becomes one `insert_parse_artifacts_batch`, and the per-file `set_file_hash` loop becomes one `set_file_hashes_batch`. Both still fire after embedding, in the same order as before — the N+1 commits go away without the write barrier moving.

Both of #60's atomicity regression tests pass, along with the rest of the suite: 771 `vera-core` tests, 92 `vera-cli`, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

The review findings from the previous round are all still addressed: the ordering test still calls the production path and still fails on a SQL-order zip, the duplicated INSERT statements are still delegated to the batch methods, and the redundant `Arc` is gone (upstream had already removed the surrounding code it wrapped).

One consequence worth flagging for review: `chunk_file_for_update`'s signature changed, dropping the `&mut Vec<FileError>` parameter in favour of returning `Option<FileError>`. It is private to the module and has two call sites, both updated, but it is a change to code #60 just touched, so it deserves a look.

citron07r:
Reviewed the rebased branch locally. Four findings: two fixed in 8606d43, two I checked and am declining with evidence.

**Redundant fixture guards (valid, fixed).** Correct — I insert the chunks in id order, so `insertion_paths` and `lexicographic_paths` are the same sequence and the second `assert_ne!` could never fail independently. Collapsed into one guard with a message that names both orders, and a comment explaining what it is protecting against.

**No direct coverage of the batch APIs (valid, added).** They only had indirect coverage through their callers. Added four tests: multi-row persistence, replace-on-repeat for `set_file_hashes_batch`, the empty-input early return for all three (an update run with nothing to write must not error), and the partial-result case for `get_chunks_by_ids` — a missing id is skipped rather than erroring, which is precisely what the caller's "chunk metadata not found, skipping" branch depends on. `insert_parse_artifacts_batch` is asserted across two files, since taking only the first file's rows is the plausible way to get that loop wrong.

**"MockProvider produces a, c, b, not c, a, b" (declined — the comment is correct).** I checked rather than assumed, by printing the order the vector store actually returns on this branch:

```
ACTUAL DISTANCE ORDER = ["src/c.rs", "src/a.rs", "src/b.rs"]
```

c, a, b, as the comment says. Worth verifying rather than reasoning about, since #60 touched `embedding/provider.rs` and could have moved it — but it did not.

**Chunking the `IN (...)` list (declined again — still not reachable).** The 32,766 limit is right, and a "result limit above 5,461" would indeed exceed it, but no such limit reaches this call. `VectorStore::search` clamps to `MAX_KNN_K = 4096` at `storage/vector.rs:217`, *before* the query runs, so the returned candidate count — and therefore `ids.len()` — is at most 4096 no matter how large the pool multipliers in `hybrid.rs` grow. `get_chunks_by_ids` still has exactly one production caller (`retrieval/vector.rs:85`), whose ids come one-per-candidate from that clamped result. 4096 against 32766 is 8x headroom.

I re-verified both the clamp and the caller list on the rebased branch rather than relying on the earlier round, since the rebase moved a lot of code. If the clamp is ever raised or a second caller appears, this becomes real and chunking is the right fix — but adding it now would be guarding a state the code cannot reach.

775 `vera-core` tests, 92 `vera-cli`, fmt clean, clippy at the 5 pre-existing warnings.

citron07r:
Rebased again — #63, #65 and #77 landed in the meantime. Only `freshness.rs` conflicted, and it is worth saying how, because the obvious resolution was wrong.

#77 changed the freshness scan so an unreadable tracked file counts as modified rather than being skipped, which is what stops a stale index looking current. This PR's parallel version returned `None` for an unreadable file and the collection loop then `continue`d past it — the pre-#77 behaviour, because this branch was written before #77 existed. Taking my side of the conflict would have reverted the fix while looking like an ordinary merge.

Verified rather than assumed, by resolving it the wrong way first:

```
freshness_scan_marks_tracked_read_failures_as_modified ... FAILED
  left: 0
 right: 1
```

The resolution keeps #77's semantics inside the parallel path: `None` still means unreadable, and it now increments `files_modified` instead of skipping. Rayon covers the read and hash; the metadata lookup stays sequential because `MetadataStore` wraps a single SQLite connection and is not `Sync`.

This is the second time this PR has conflicted semantically rather than textually — the same thing happened with #60's deferred-write restructure. Both times the compile-clean merge would have quietly undone someone else's fix, and both times the upstream PR's own regression test was what proved it.

All three of the relevant upstream regression tests pass on the rebased branch: `freshness_scan_marks_tracked_read_failures_as_modified` (#77), plus `update_embedding_failure_preserves_existing_parse_data` and `update_keeps_indexed_data_when_a_discovered_file_cannot_be_read` (#60).

777 `vera-core` tests, 97 `vera-cli`, `cargo fmt --check` clean, clippy at the 5 pre-existing warnings.

citron07r:
Three findings I had not answered. Two were weak assertions and are fixed in 620405e; the third I am declining with a reason.

**Both test assertions were weaker than their names (fixed).**

`insert_parse_artifacts_batch_persists_both_kinds_across_files` asserted only that `find_callers("helper")` returned 2 rows. Both could have been stored under the same path, so it did not verify the "across files" in its own name. It now asserts the returned `CallerRef.file_path` values equal exactly `["src/lib.py", "src/main.rs"]`. Confirmed it catches the gap: repointing the second entry at `src/main.rs` yields

```
left:  ["src/main.rs", "src/main.rs"]
right: ["src/lib.py", "src/main.rs"]
```

while the old count check stayed green under that same mutation.

`set_file_hashes_batch_persists_every_row_and_replaces_on_repeat` re-checked only `src/main.rs` after the second batch call, so a batch that clobbered untouched rows would have passed. It now asserts `src/lib.py` still reads `hash-b`. Verified by injecting the actual defect, a `DELETE FROM file_hashes` inside the batch transaction:

```
left:  None
right: Some("hash-b")
```

The pre-existing `hash-c` assertion passed against that clear-then-write batch, so the test was fully green on a genuinely broken implementation.

**Cloning in the single-item wrappers (declined).** Correct that `insert_references` and `insert_type_relations` now `to_vec()` the slice before delegating, where the direct inserts did not.

The reason I am leaving it: after this PR those three single-item methods have **no production callers**. `pipeline.rs` and `update.rs` both go through the batch APIs; the only remaining callers are tests (`metadata.rs`, `freshness.rs`, `graph_augmentation.rs`). So the allocation is confined to test setup, and adding a borrowed insertion helper would mean a third code path carrying the same SQL, which is what the deduplication these findings came from was removing.

The cleaner resolution is the one I flagged when the duplication was first raised: delete the single-item methods outright, since nothing in production calls them. I did not do that here because removing public API felt like a maintainer's call rather than something a perf PR should decide. If you would prefer that, say so and it is a small follow-up.

777 `vera-core` tests, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

PR #80, opened 2026-08-19
perf(storage): index the lower() lookups instead of scanning

### LINKED ISSUE #78: lower() predicates make four indexes unusable; vera dead-code takes 4.4s on a 4790-chunk index
## Summary

Five hot queries wrap an indexed column in `lower()`. SQLite cannot use an index on a bare column to satisfy a predicate over an expression, so each of these full-scans its table. The indexes that were created to serve them (`idx_refs_callee`, `idx_refs_caller`, `idx_type_relations_target`, `idx_chunks_symbol_name`) are never used for seeks — they are maintained on every write and serve no read.

The visible symptom is `vera dead-code`, which on a small index takes **4.4 seconds**.

## Affected queries

All in `crates/vera-core/src/storage/metadata.rs`:

| Method | Line | Predicate |
|---|---|---|
| `get_chunks_by_symbol_name` | 320 | `lower(symbol_name) = lower(?1)` |
| `find_callers` | 660 | `lower(callee) = lower(?1)` |
| `find_type_relations` | 683 | `lower(target) = lower(?1)` |
| `find_callees` | 713 | `lower(caller) = lower(?1)` |
| `find_dead_symbols` | 744 | `lower(r.callee) = lower(c.symbol_name)` inside a correlated `NOT EXISTS` |

These are not cold paths. `get_chunks_by_symbol_name` runs on every identifier-shaped query, and graph augmentation issues `find_callers` and `find_type_relations` up to five times each per search.

`find_dead_symbols` is the worst case: the correlated `NOT EXISTS` re-scans `[references]` once per candidate chunk, so it is O(chunks × references).

## Reproduction

Against Vera's own index of this repository (4790 chunks, 28118 references, 1033 type relations), using the real `metadata.db`:

```
sqlite> EXPLAIN QUERY PLAN SELECT file_path, line, caller FROM [references]
   ...> WHERE lower(callee) = lower('search_vector_with_stores');
`--SCAN references

sqlite> EXPLAIN QUERY PLAN SELECT id, file_path FROM chunks
   ...> WHERE lower(symbol_name) = lower('build_session');
`--SCAN chunks

sqlite> EXPLAIN QUERY PLAN SELECT file_path, line, owner, target, kind FROM type_relations
   ...> WHERE lower(target) = lower('Display');
`--SCAN type_relations
```

`find_dead_symbols` before:

```
|--SCAN c
`--CORRELATED SCALAR SUBQUERY 1
   `--SCAN r USING COVERING INDEX idx_refs_callee
Run Time: real 4.395s
```

Note the subquery *scans* `idx_refs_callee` rather than seeking it — that index's only role today is to make the full scan slightly cheaper.

## Effect of expression indexes

Adding `CREATE INDEX ... ON [references](lower(callee))` and `ON chunks(lower(symbol_name))` to the same database:

```
`--SEARCH references USING INDEX idx_refs_callee_lower (<expr>=?)
`--SEARCH chunks USING INDEX idx_chunks_symbol_name_lower (<expr>=?)
```

| Query | Before | After |
|---|---|---|
| `find_callers` × 200 | 0.390 s | 0.00018 s |
| `get_chunks_by_symbol_name` × 200 | 0.063 s | 0.00006 s |
| `find_dead_symbols` × 1 | 4.395 s | 0.002 s |

`find_dead_symbols` returns the same 440 rows before and after, so this is a plan change, not a semantic one.

## Proposed fix

Add four expression indexes in `init_schema` alongside the existing ones, with `CREATE INDEX IF NOT EXISTS` so existing databases pick them up on next open:

- `[references](lower(callee))`
- `[references](lower(caller))`
- `type_relations(lower(target))`
- `chunks(lower(symbol_name))`

`lower()` is deterministic, so it is legal in an index. The index expression has to match the query text, which the EXPLAIN output above confirms.

Two follow-on questions for maintainers, which I have deliberately not folded into the fix:

1. Once these land, the four bare-column indexes (`idx_refs_callee`, `idx_refs_caller`, `idx_type_relations_target`, `idx_chunks_symbol_name`) are provably unused for seeks and are pure write amplification on the two largest tables. Dropping them is a separate migration and worth deciding on its own.
2. The alternative design is to stop lower-casing in SQL and instead store a normalised column, or use `COLLATE NOCASE` on the column definition. That is a schema change with a migration, and it changes behaviour for non-ASCII identifiers, so it is a bigger decision than an added index. Expression indexes get the win without touching semantics.

Happy to send the PR for the indexes.

citron07r:
## Problem

Five queries in `metadata.rs` match on `lower(column)`. SQLite cannot use an index on a bare column to satisfy a predicate over an expression, so every one of them full-scans its table — and the four indexes created to serve them (`idx_refs_callee`, `idx_refs_caller`, `idx_type_relations_target`, `idx_chunks_symbol_name`) are never used for seeks. They are maintained on every write and serve no read.

`find_dead_symbols` is the worst case. Its correlated `NOT EXISTS` re-scans `[references]` once per candidate chunk, so the cost is O(chunks × references).

## Change

Four expression indexes in `init_schema`, matching the query text:

- `[references](lower(callee))`
- `[references](lower(caller))`
- `type_relations(lower(target))`
- `chunks(lower(symbol_name))`

`lower()` is deterministic, so it is legal in an index. `CREATE INDEX IF NOT EXISTS` means existing databases pick them up on the next open without a migration step.

No query text changed, so there is no behaviour change to reason about — only which plan SQLite picks.

## Verification

Reproduced against this repository's own index (4790 chunks, 28118 references, 1033 type relations), with the current released binary as the control on the same index and the same query.

Plans, before and after:

```
find_callers               SCAN references     ->  SEARCH references USING INDEX idx_refs_callee_lower (<expr>=?)
get_chunks_by_symbol_name  SCAN chunks         ->  SEARCH chunks USING INDEX idx_chunks_symbol_name_lower (<expr>=?)
find_type_relations        SCAN type_relations ->  SEARCH type_relations USING INDEX idx_type_relations_target_lower (<expr>=?)
find_dead_symbols          SCAN r USING COVERING INDEX idx_refs_callee
                                               ->  SEARCH r USING COVERING INDEX idx_refs_callee_lower (<expr>=?)
```

End to end, same index, same command:

| Command | Control (master) | This branch | Output |
|---|---|---|---|
| `vera dead-code` | 3.622 s | **0.010 s** | 443 lines, byte-identical |
| `vera references build_session` | 0.020 s | **0.008 s** | identical |

The first `dead-code` run against a pre-existing index takes 1.714 s while SQLite builds the four indexes, which is still below the control; every run after that is the 0.010 s figure.

Raw SQL timings on the same database, 200 iterations each: `find_callers` 0.390 s → 0.00018 s, `get_chunks_by_symbol_name` 0.063 s → 0.00006 s. `find_dead_symbols` returns the same 440 rows in both cases, confirming this is a plan change rather than a semantic one.

## Tests

Two tests that assert on `EXPLAIN QUERY PLAN` rather than on results. That is deliberate: the index expression has to match the query text character for character, and if either side is edited without the other, SQLite silently falls back to a full scan and no result-based test would notice.

Confirmed they catch it — with the four indexes removed:

```
find_type_relations still scans its table: SCAN type_relations USING INDEX idx_type_relations_file_path | ...
the reference subquery must seek, not scan: ... | SCAN r USING COVERING INDEX idx_refs_callee
```

772 `vera-core` tests pass, `cargo fmt --check` clean, clippy unchanged at the pre-existing warnings.

## Deliberately not included

Two follow-ups that belong to maintainers rather than this PR:

1. Once these land, the four bare-column indexes are provably unused for seeks and are pure write amplification on the two largest tables. Dropping them is a schema migration against existing user databases and deserves its own decision.
2. The alternative design is `COLLATE NOCASE` on the column, or storing a normalised column. Both are schema changes with migrations, and both alter behaviour for non-ASCII identifiers. Expression indexes get the win without touching semantics, so they seemed the right first step.

Fixes #78

---
## Summary by cubic
Indexes the lower()-based metadata lookups so SQLite seeks instead of scanning, and strengthens plan tests to prevent silent regressions. Before, predicates on lower(column) full-scanned `references`, `chunks`, and `type_relations`; now four expression indexes match the query text, yielding seeks with no semantic changes.

- Adds indexes: `[references](lower(callee))`, `[references](lower(caller))`, `type_relations(lower(target))`, `chunks(lower(symbol_name))`.
- Extracts the lookup SQL into shared constants; plan tests assert the exact expression index is used and reject any SCAN.
- Query text is unchanged; schema init adds indexes via `CREATE INDEX IF NOT EXISTS`. Bare-column indexes remain and are currently unused for these seeks.

**Rollout**
- No migration required; existing databases build the new indexes on next open.
- First run may take ~1–2s to build indexes; subsequent runs use seeks.

<sup>Written for commit ca83719cbed04029f6e9d7e6b8b352932c3769dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/80?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved case-insensitive searches for symbols, callers, callees, type relations, and dead-symbol references.
  * Added indexes to accelerate matching and reduce unnecessary database scans.
  * Optimized query execution for faster search results.

### DIFF
--- a/crates/vera-core/src/storage/metadata.rs
+++ b/crates/vera-core/src/storage/metadata.rs
@@ -103,6 +103,44 @@ pub struct LanguageHealthStat {
     pub files_with_parse_failures: u64,
 }
 
+/// Case-insensitive lookup queries, held as constants so the query-plan tests
+/// assert against the exact text the production methods run.
+///
+/// An expression index only applies when the index expression matches the
+/// query text. A test that asserted on its own copy of the SQL would keep
+/// passing after the original drifted, and the planner would silently fall
+/// back to a full scan with nothing failing.
+const SQL_CHUNKS_BY_SYMBOL_NAME: &str =
+    "SELECT id, file_path, line_start, line_end, content, language,
+                        symbol_type, symbol_name
+                 FROM chunks
+                 WHERE lower(symbol_name) = lower(?1)
+                 ORDER BY file_path, line_start";
+
+const SQL_FIND_CALLERS: &str = "SELECT file_path, line, caller FROM [references]
+                 WHERE lower(callee) = lower(?1)
+                 ORDER BY file_path, line";
+
+const SQL_FIND_CALLEES: &str = "SELECT file_path, line, callee FROM [references]
+                 WHERE lower(caller) = lower(?1)
+                 ORDER BY file_path, line";
+
+const SQL_FIND_TYPE_RELATIONS: &str = "SELECT file_path, line, owner, target, kind
+                 FROM type_relations
+                 WHERE lower(target) = lower(?1)
+                 ORDER BY file_path, line, owner";
+
+const SQL_FIND_DEAD_SYMBOLS: &str = "SELECT c.symbol_name, c.file_path, c.line_start, c.symbol_type
+                 FROM chunks c
+                 WHERE c.symbol_name IS NOT NULL
+                   AND c.symbol_type IN ('function', 'method')
+                   AND lower(c.symbol_name) NOT IN ('main', 'new', 'default', 'drop', 'clone', 'fmt', 'from', 'into', 'deref', 'init', 'setup', 'teardown')
+                   AND NOT EXISTS (
+                       SELECT 1 FROM [references] r
+                       WHERE lower(r.callee) = lower(c.symbol_name)
+                   )
+                 ORDER BY c.file_path, c.line_start";
+
 /// SQLite-backed metadata store for chunk attributes.
 pub struct MetadataStore {
     conn: Connection,
@@ -329,13 +367,7 @@ impl MetadataStore {
     pub fn get_chunks_by_symbol_name(&self, symbol_name: &str) -> Result<Vec<Chunk>> {
         let mut stmt = self
             .conn
-            .prepare_cached(
-                "SELECT id, file_path, line_start, line_end, content, language,
-                        symbol_type, symbol_name
-                 FROM chunks
-                 WHERE lower(symbol_name) = lower(?1)
-                 ORDER BY file_path, line_start",
-            )
+            .prepare_cached(SQL_CHUNKS_BY_SYMBOL_NAME)
             .context("failed to prepare symbol chunks query")?;
 
         let rows = stmt
@@ -671,11 +703,7 @@ impl MetadataStore {
     pub fn find_callers(&self, symbol_name: &str) -> Result<Vec<CallerRef>> {
         let mut stmt = self
             .conn
-            .prepare_cached(
-                "SELECT file_path, line, caller FROM [references]
-                 WHERE lower(callee) = lower(?1)
-                 ORDER BY file_path, line",
-            )
+            .prepare_cached(SQL_FIND_CALLERS)
             .context("failed to prepare callers query")?;
         let rows = stmt
             .query_map(params![symbol_name], |row| {
@@ -693,12 +721,7 @@ impl MetadataStore {
     pub fn find_type_relations(&self, symbol_name: &str) -> Result<Vec<TypeRelationRef>> {
         let mut stmt = self
             .conn
-            .prepare_cached(
-                "SELECT file_path, line, owner, target, kind
-                 FROM type_relations
-                 WHERE lower(target) = lower(?1)
-                 ORDER BY file_path, line, owner",
-            )
+            .prepare_cached(SQL_FIND_TYPE_RELATIONS)
             .context("failed to prepare type relation query")?;
         let rows = stmt
             .query_map(params![symbol_name], |row| {
@@ -724,11 +747,7 @@ impl MetadataStore {
     pub fn find_callees(&self, symbol_name: &str) -> Result<Vec<CalleeRef>> {
         let mut stmt = self
             .conn
-            .prepare_cached(
-                "SELECT file_path, line, callee FROM [references]
-                 WHERE lower(caller) = lower(?1)
-                 ORDER BY file_path, line",
-            )
+            .prepare_cached(SQL_FIND_CALLEES)
             .context("failed to prepare callees query")?;
         let rows = stmt
             .query_map(params![symbol_name], |row| {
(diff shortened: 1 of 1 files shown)


citron07r:
Good finding, and acting on it exposed a second problem underneath. Both fixed in de858b6.

**Tests owned copies of the production SQL (valid, fixed).** Exactly right, and the copies had already drifted — as noted, `find_type_relations` orders by `file_path, line, owner` while my copy stopped at `line`, and the chunk query's `ORDER BY` was missing altogether. That is the whole failure mode this test was written to prevent, reproduced inside the test itself: production drifts out of matching its index expression, the planner silently falls back to a scan, and a test asserting on a stale copy keeps passing.

Each of the five queries is now a `const` used by both the method and its test, so they cannot diverge.

**The assertion was also too weak to catch it (found while verifying the fix).** After extracting the constants I checked the test could now detect production drift, by editing the production query to `lower(trim(callee))` so it no longer matches the index. It still passed.

The reason is worth recording: the assertion looked for `"USING INDEX"` in the plan, but a full scan can read *through* an index and report

```
SCAN references USING INDEX idx_refs_file_path | USE TEMP B-TREE FOR LAST TERM OF ORDER BY
```

which contains `USING INDEX`. The check passed on precisely the plan it existed to reject. The negative half was broken too — it looked for `SCAN [references]` with the bracket-quoted table name, but EXPLAIN prints `SCAN references`.

It now requires the plan to start with `SEARCH` and rejects any `SCAN`, which is the distinction that actually matters.

Verified in both directions after the change: the `lower(trim(callee))` edit now fails, and removing the four indexes still fails. So the test catches drift on either side of the index/query pair, which it did not before.

End-to-end unchanged: `vera dead-code` output is byte-identical to the released binary at 443 lines, in 0.012 s. 772 `vera-core` tests, 92 `vera-cli`, fmt clean, clippy at the 5 pre-existing warnings.

PR #81, opened 2026-08-19
perf(storage): seek the index when deleting a file's vectors

### LINKED ISSUE #79: delete_by_file_prefix full-scans chunk_id_map (LIKE ... ESCAPE defeats the index), commits per row, and the table carries a duplicate index
## Summary

`VectorStore::delete_by_file_prefix` (`crates/vera-core/src/storage/vector.rs:293`) has three separate problems on a path that runs once per changed file during every incremental update:

1. The `LIKE ?1 ESCAPE '\'` predicate disables SQLite's LIKE-prefix optimization, so it full-scans `chunk_id_map`.
2. The deletion loop issues two uncached `Connection::execute` calls per matched rowid with no enclosing transaction, so every row is its own commit.
3. `chunk_id_map` carries a redundant index: `chunk_id` is declared `UNIQUE`, which already creates `sqlite_autoindex_chunk_id_map_1`, and `idx_chunk_id_map` is then created on the same column.

## Reproduction

Against Vera's own `vectors.db` for this repository (4790 rows):

```
sqlite> EXPLAIN QUERY PLAN SELECT rowid, chunk_id FROM chunk_id_map
   ...> WHERE chunk_id LIKE 'crates/vera-core/src/retrieval/%' ESCAPE '\';
`--SCAN chunk_id_map
```

SQLite will convert `LIKE 'prefix%'` into a range scan when the column is indexed, but the `ESCAPE` clause disqualifies the optimization outright, so the index is never consulted. Rewritten as an explicit range:

```
sqlite> EXPLAIN QUERY PLAN SELECT rowid, chunk_id FROM chunk_id_map
   ...> WHERE chunk_id >= 'crates/vera-core/src/retrieval/'
   ...>   AND chunk_id <  'crates/vera-core/src/retrieval0';
`--SEARCH chunk_id_map USING COVERING INDEX idx_chunk_id_map (chunk_id>? AND chunk_id<?)
```

300 prefix lookups: **0.023 s → 0.003 s**, identical result counts. The gap widens with table size, since one form is O(rows) and the other O(log n + matches).

On the redundant index, both the equality lookup and the range scan are served by the UNIQUE constraint's automatic index once the explicit one is removed:

```
sqlite> DROP INDEX idx_chunk_id_map;
sqlite> EXPLAIN QUERY PLAN SELECT rowid, chunk_id FROM chunk_id_map WHERE chunk_id >= 'crates/' AND chunk_id < 'crates0';
`--SEARCH chunk_id_map USING COVERING INDEX sqlite_autoindex_chunk_id_map_1 (chunk_id>? AND chunk_id<?)

sqlite> EXPLAIN QUERY PLAN SELECT rowid FROM chunk_id_map WHERE chunk_id = 'x';
`--SEARCH chunk_id_map USING COVERING INDEX sqlite_autoindex_chunk_id_map_1 (chunk_id=?)
```

So `idx_chunk_id_map` is pure write amplification on every vector insert.

## Proposed fix

- Replace the `LIKE ... ESCAPE` with an explicit half-open range. This also removes the need to escape `%` and `_` at all, since a range predicate has no wildcards — the existing test covering paths that contain those characters becomes a test of the range bound instead.
- Wrap the deletion loop in a single transaction and use `prepare_cached`.
- Drop `idx_chunk_id_map`.

The one thing needing care is deriving the upper bound for an arbitrary UTF-8 prefix. Incrementing the last byte is wrong in general; the bound has to be computed on the last character, with a fallback for the case where it cannot be incremented.

I have this working locally and will send a PR. Filing separately because the redundant-index removal is a schema migration against existing user databases and a maintainer may want to take that part on its own timeline.

citron07r:
## Problem

`VectorStore::delete_by_file_prefix` runs once per changed file on every incremental update, and had three independent problems.

**1. The predicate could not use the index.** SQLite converts `LIKE 'prefix%'` into a range scan when the column is indexed, but the `ESCAPE` clause disqualifies that optimization outright:

```
sqlite> EXPLAIN QUERY PLAN SELECT rowid, chunk_id FROM chunk_id_map
   ...> WHERE chunk_id LIKE 'crates/vera-core/src/retrieval/%' ESCAPE '\';
`--SCAN chunk_id_map
```

**2. Every row was its own commit.** The deletion loop issued two uncached `Connection::execute` calls per matched rowid with no enclosing transaction.

**3. The table carried a duplicate index.** `chunk_id` is declared `UNIQUE`, which already builds `sqlite_autoindex_chunk_id_map_1` over that column; `idx_chunk_id_map` was then created on the same column — a second B-tree maintained on every vector insert for nothing.

## Change

A half-open range instead of `LIKE ... ESCAPE`, one transaction with cached statements for the deletes, and the redundant index dropped.

The range form also removes the wildcard-escaping logic entirely, because a range predicate has no wildcards to confuse. That is a simplification rather than a behaviour change — the escaping existed only to defend against the `LIKE` semantics.

```
`--SEARCH chunk_id_map USING COVERING INDEX sqlite_autoindex_chunk_id_map_1 (chunk_id>? AND chunk_id<?)
```

## Verification

Against this repository's own `vectors.db` (4790 rows), driving 300 real chunk-id prefixes taken from the table itself:

| | Rows returned | Time |
|---|---|---|
| `LIKE ... ESCAPE` | 19674 | 0.0477 s |
| range | 19674 | 0.0034 s |

Identical results, ~14x faster. The gap widens with table size, since one form is O(rows) and the other O(log n + matches).

On the duplicate index, both access patterns are served by the UNIQUE constraint's automatic index once the explicit one is gone:

```
sqlite> DROP INDEX idx_chunk_id_map;
sqlite> EXPLAIN QUERY PLAN SELECT rowid FROM chunk_id_map WHERE chunk_id = 'x';
`--SEARCH chunk_id_map USING COVERING INDEX sqlite_autoindex_chunk_id_map_1 (chunk_id=?)
```

End to end on a copy of a real index, the migration runs on open and search keeps working:

```
indexes before:  sqlite_autoindex_chunk_id_map_1, idx_chunk_id_map
$ vera search "vector similarity search"
crates/vera-cli/src/commands/search.rs:77-85 struct:SearchRunner
indexes after:   sqlite_autoindex_chunk_id_map_1
```

777 `vera-core` tests, 92 `vera-cli`, `cargo fmt --check` clean, clippy unchanged at the pre-existing warnings.

## The one subtle part

Deriving the exclusive upper bound for an arbitrary UTF-8 prefix. Two cases are easy to get wrong: the successor of a code point can land in the UTF-16 surrogate range, which is not a valid `char`; and a trailing `char::MAX` has no successor at all, so the character before it has to be incremented and the max dropped. `prefix_upper_bound` handles both and returns `None` when no bound exists, in which case `chunk_id >= prefix` is already exact.

This is unit-tested directly rather than only through `delete_by_file_prefix`, because the end-to-end tests cannot distinguish a bound that is off by one character — I checked, by introducing exactly that bug: the unit test caught it, the integration tests did not.

Test coverage added: wildcard-looking paths (`src/a_b.rs`, `src/100%.rs`) must not over-match their literal siblings; the prefix boundary must exclude `src/app2.rs` while including every descendant of `src/app.rs:`; non-ASCII paths; the awkward code points above; and a plan assertion that the range actually seeks the index, since that is the entire point and a result-based test would pass either way.

Worth noting the pre-existing `delete_by_file_prefix` test did not cover the wildcard cases, so the escaping behaviour was previously unpinned.

Fixes #79

---
## Summary by cubic
Speeds up and hardens file‑prefix deletions by replacing a `LIKE ... ESCAPE` full scan with an indexed half‑open range and running the scan and deletes inside a single IMMEDIATE transaction. This is ~10–15x faster on real data and closes a race where inserts between scan and delete could survive.

- Review
  - Prefix match uses `chunk_id >= prefix` and, when available, `chunk_id < upper_bound(prefix)`; when no bound exists, `>= prefix` is exact. `prefix_upper_bound` skips UTF‑16 surrogates and handles trailing `char::MAX`.
  - The UNIQUE `chunk_id` autoindex serves equality and range; EXPLAIN confirms an index seek (plan test added).
  - One IMMEDIATE transaction opens before the scan; tests assert a contending writer fails at BEGIN and that the transaction later releases for subsequent writes.
  - Deletes from `vec_chunks` and `chunk_id_map` use cached statements and the same rowids via `rowids_with_prefix`. Tests cover literal `%`/`_`, non‑ASCII paths, and prefix boundaries.

- Rollout
  - On open, drop `idx_chunk_id_map`. No action required; the autoindex remains.

<sup>Written for commit 8fc83c6a3ba053e5a78d36fd3214a918dcd54d1f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/81?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved prefix-based deletion for paths containing special characters, wildcard literals, non-ASCII text, and boundary cases.
  - Corrected deletion behavior for prefixes without a defined upper limit.
  - Improved handling of empty prefixes and maximum-character boundaries.
  - Ensured related path data is removed consistently and atomically during deletion.

- **Performance**
  - Optimized prefix deletions for faster, more efficient cleanup.
  - Reduced unnecessary storage overhead during initialization.

### DIFF
--- a/crates/vera-core/src/storage/vector.rs
+++ b/crates/vera-core/src/storage/vector.rs
@@ -8,6 +8,10 @@ use rusqlite::{Connection, OptionalExtension, ffi::sqlite3_auto_extension, param
 use sqlite_vec::sqlite3_vec_init;
 use zerocopy::IntoBytes;
 
+const PREFIX_RANGE_SQL: &str =
+    "SELECT rowid FROM chunk_id_map WHERE chunk_id >= ?1 AND chunk_id < ?2";
+const PREFIX_LOWER_BOUND_SQL: &str = "SELECT rowid FROM chunk_id_map WHERE chunk_id >= ?1";
+
 /// sqlite-vec backed vector store for embedding search.
 pub struct VectorStore {
     conn: Connection,
@@ -377,14 +381,8 @@ fn rowids_with_prefix(conn: &Connection, prefix: &str) -> Result<Vec<i64>> {
     // SQL would cost the index range plan, which is the point of all this.
     let upper = prefix_upper_bound(prefix);
     let (sql, args): (&str, Vec<&dyn rusqlite::ToSql>) = match &upper {
-        Some(upper) => (
-            "SELECT rowid FROM chunk_id_map WHERE chunk_id >= ?1 AND chunk_id < ?2",
-            vec![&prefix, upper],
-        ),
-        None => (
-            "SELECT rowid FROM chunk_id_map WHERE chunk_id >= ?1",
-            vec![&prefix],
-        ),
+        Some(upper) => (PREFIX_RANGE_SQL, vec![&prefix, upper]),
+        None => (PREFIX_LOWER_BOUND_SQL, vec![&prefix]),
     };
 
     let mut stmt = conn
@@ -756,11 +754,9 @@ mod tests {
     }
 
     #[test]
-    fn prefix_delete_scans_and_deletes_in_one_transaction() {
-        // The scan runs inside the transaction, so a second connection cannot
-        // insert a matching row between finding the rowids and deleting them.
-        // An IMMEDIATE transaction takes the write lock up front, which is
-        // what makes the competing writer wait rather than slip in.
+    fn prefix_delete_releases_transaction_for_subsequent_writes() {
+        // A second connection can write after the prefix delete commits, which
+        // confirms that the transaction was committed and released.
         let dir = tempfile::tempdir().unwrap();
         let path = dir.path().join("vectors.db");
         let store = VectorStore::open(&path, 4).unwrap();
@@ -789,10 +785,9 @@ mod tests {
         let store = VectorStore::open_in_memory(4).unwrap();
         let mut stmt = store
             .conn
-            .prepare(
-                "EXPLAIN QUERY PLAN
-                 SELECT rowid FROM chunk_id_map WHERE chunk_id >= ?1 AND chunk_id < ?2",
-            )
+            .prepare(&format!(
+                "EXPLAIN QUERY PLAN\n                 {PREFIX_RANGE_SQL}"
+            ))
             .unwrap();
         let plan: Vec<String> = stmt
             .query_map(params!["a", "b"], |row| row.get::<_, String>(3))
@@ -801,8 +796,9 @@ mod tests {
             .collect();
         let plan = plan.join(" | ");
         assert!(
-            plan.contains("USING COVERING INDEX") || plan.contains("USING INDEX"),
-            "prefix range must seek the index: {plan}"
+            plan.contains("SEARCH")
+                && (plan.contains("USING COVERING INDEX") || plan.contains("USING INDEX")),
+            "prefix range must search using an index: {plan}"
         );
     }
 }
--- a/crates/vera-core/src/storage/vector.rs
+++ b/crates/vera-core/src/storage/vector.rs
@@ -310,18 +310,27 @@ impl VectorStore {
         // form scans `chunk_id_map` in full where a range seeks the index. It
         // also removes the need to escape `%` and `_`, since a range has no
         // wildcards to confuse.
-        let rows = self.rowids_with_prefix(prefix)?;
+        // One transaction for the whole file instead of two commits per row,
+        // opened *before* the scan so a concurrent writer on the same database
+        // cannot insert a matching row between finding the rowids and deleting
+        // them — that row would survive while this reported success.
+        //
+        // IMMEDIATE rather than the default DEFERRED because this transaction
+        // reads and then writes: a deferred transaction starts as a reader and
+        // can fail the upgrade with SQLITE_BUSY_SNAPSHOT, which `busy_timeout`
+        // does not retry.
+        let tx = rusqlite::Transaction::new_unchecked(
+            &self.conn,
+            rusqlite::TransactionBehavior::Immediate,
+        )
+        .context("failed to begin prefix delete transaction")?;
 
+        let rows = rowids_with_prefix(&tx, prefix)?;
         let count = rows.len() as u64;
         if count == 0 {
             return Ok(count);
         }
 
-        // One transaction for the whole file instead of two commits per row.
-        let tx = self
-            .conn
-            .unchecked_transaction()
-            .context("failed to begin prefix delete transaction")?;
         {
             let mut delete_vector = tx
                 .prepare_cached("DELETE FROM vec_chunks WHERE rowid = ?1")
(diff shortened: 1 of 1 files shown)


citron07r:
Both findings valid, both fixed in 0de9723.

**Scan ran outside the transaction (fixed).** Correct: the rowid scan happened before the transaction opened, so a second connection on the same database could insert a matching row in the gap, and it would survive while the method reported success. `vera watch` and `vera update` on the same index are exactly that shape.

Worth being straight about provenance: the race is older than this PR. The previous `LIKE`-based version had the same scan-then-delete gap and no transaction at all, so this PR narrowed the window rather than opening it. That is not a reason to leave it — this is the code being rewritten, so it is the right moment to close it properly.

One thing I changed beyond the suggestion: the transaction is **IMMEDIATE**, not the default DEFERRED. Moving the scan inside makes this a read-then-write transaction, which is the classic case where a deferred transaction starts as a reader and then fails the upgrade with `SQLITE_BUSY_SNAPSHOT` — and SQLite returns that directly rather than invoking the busy handler, so `busy_timeout` cannot retry it. Taking the write lock up front avoids trading a lost-row race for an intermittent BUSY under the same concurrency. `Transaction::new_unchecked` accepts a behaviour and `&Connection`, so this needs no `&mut self`.

Added a test covering the two-connection case and asserting the transaction is released rather than left open.

**Duplicated rowid-query logic (fixed).** Folded into one helper, with the caveat you flagged respected: the two predicates stay distinct, because merging them into a single statement would cost the index range plan that is the entire point of the change. Only the execution and collection are shared. The helper now takes a `&Connection` rather than `&self`, which is also what lets the caller hand it the open transaction — so the two fixes turned out to be the same refactor.

Re-verified the plan still seeks after moving the scan inside the transaction, since that was the thing most likely to regress silently: 300 real prefixes against this repository's own `vectors.db`, 19674 rows in 0.0032 s, and `sqlite_master` confirms only `sqlite_autoindex_chunk_id_map_1` remains after the migration.

778 `vera-core` tests, 92 `vera-cli`, `cargo fmt --check` clean, clippy at the 5 pre-existing warnings.

citron07r:
Both findings were valid — both were tests of mine asserting something adjacent to the property rather than the property itself. @lemon07r had already pushed 516e6a9 covering the first while I was working on it; I have rebased onto that rather than overwriting it, and added the piece it left open.

**Plan test owned a copy of the SQL (fixed in 516e6a9).** Right, and it is the same defect CodeRabbit caught in the plan tests on #80: the test built its own hard-coded range query, so reverting the helper to `LIKE` would have left it green. 516e6a9's `PREFIX_RANGE_SQL` / `PREFIX_LOWER_BOUND_SQL` constants are the right fix and its assertion (`SEARCH` *and* an index) is stricter than what I had written, so I have kept it as-is. I did verify it now catches the regression: swapping the production SQL back to `LIKE ... ESCAPE` fails it with `SCAN chunk_id_map`.

**Concurrency test never overlapped the connections (addressed in 082335f).** Also right, and worth separating into two claims. 516e6a9 renamed the test to `prefix_delete_releases_transaction_for_subsequent_writes`, which is an honest description of what it verified — the transaction is committed and released. That is a real property and worth keeping.

But it leaves the property the IMMEDIATE-before-scan ordering actually exists for unasserted: that a competing writer cannot slip a matching row in between the scan and the deletes. So I have added a second test rather than changing the renamed one.

It turns out that ordering is observable without thread timing, which avoids a flaky test. With a competitor holding the write lock:

- taking an IMMEDIATE transaction up front fails at `BEGIN`, before any scanning;
- scanning first and opening a DEFERRED transaction afterwards gets as far as the first `DELETE`.

Asserting *which step* failed pins the ordering deterministically. Confirmed by reverting to scan-then-deferred, which fails the test with `failed to delete vector by prefix` instead of the BEGIN context.

One practical note: `busy_timeout=0` on the contended connection, since rusqlite defaults to five seconds and this test wants the lock to be unavailable immediately — without it the test spent 5.2s waiting.

779 `vera-core` tests, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

PR #84, opened 2026-08-19
perf(indexing): delete BM25 documents for all changed files in one pass

### LINKED ISSUE #82: BM25 delete allocates a 50MB IndexWriter and joins merge threads once per changed file (~26ms each)
## Summary

`Bm25Index::delete_by_file` (`crates/vera-core/src/storage/bm25.rs:250`) allocates a fresh `IndexWriter` with a 50 MB heap, commits, and blocks on `wait_merging_threads()` — and it is called once per changed file, from the modified-files loop in `update_repository_with_options_and_progress` (`crates/vera-core/src/indexing/update.rs:546` via `remove_file_chunk_data`).

The per-file cost is the writer lifecycle, not the deletion itself. Deleting one term is trivial; allocating the writer, committing a segment, and joining the merge threads is not.

## Measurement

200 documents indexed, then 25 files deleted one at a time through the current `delete_by_file`:

```
25 per-file deletes took 656.84ms (26.27ms each)
```

**~26 ms per changed file**, essentially independent of how much is being deleted. An incremental update touching 200 files therefore spends roughly **5 seconds** on writer allocation and merge barriers alone, before any parsing or embedding work.

This is measured with a throwaway probe against `Bm25Index` directly, not inferred from reading the code.

## Why it is worth fixing

This sits on the incremental update path, which is the common case — `vera watch` and `vera update` both go through it. It scales linearly with the number of changed files, so it is worst exactly when a user has done a large refactor or switched branches.

## Proposed fix

Add a batch entry point that takes all the paths and uses a single writer:

```
delete_by_files(&[&str]) -> one writer, delete_term per path, one commit, one merge join
```

The complication is that `remove_file_chunk_data` currently interleaves the three stores per file — vector prefix delete, then BM25, then chunk metadata — so batching the BM25 half means splitting the deletion phase into per-store passes. That needs care to keep the existing ordering, and it is why I am filing this rather than sending a drive-by patch.

`delete_by_chunk_id` (`bm25.rs:233`) has the same shape and the same fix, though it is not on a per-file loop today.

## Related

Same update path as #69. Independent of #78 and #79, which are SQLite-side; this one is tantivy-side and, at 26 ms per file, is plausibly the larger remaining cost on that path.

citron07r:
## Problem

`Bm25Index::delete_by_file` allocates a `WRITER_HEAP_SIZE` (50 MB) `IndexWriter`, commits a segment, and blocks on `wait_merging_threads()`. That is the writer lifecycle, not the deletion — it costs roughly the same whether one term is removed or a thousand — and it was paid **once per changed file**, from both the deleted-files loop and the modified-files loop in `update_repository_with_options_and_progress`.

## Change

A batch entry point using one writer for all paths. `delete_by_file` delegates to it, so single-file callers are unchanged and there is one implementation.

The update path now collects the paths from both loops and issues a single delete.

## Measurement

200-document index, deleting N files each way. The doc counts of the two resulting indexes are asserted equal, so this compares like with like:

| Files deleted | Per-file | Batch |
|---|---|---|
| 10 | 216 ms | 32 ms |
| 50 | 1.190 s | 29 ms |
| 100 | 2.482 s | 37 ms |

The batch stays flat at ~30 ms because it is one writer lifecycle regardless of N, so the saving grows linearly with how many files changed. At 100 files that is **67x**, and it is worst exactly when a user has done a large refactor or switched branches.

## Why the restructure is safe

The three per-file store deletions turned out to be independent: the vector delete keys on a path prefix, the BM25 delete on the path, and the metadata delete on the path, with no value passed between them. The `get_chunks_by_file` call at the top of `remove_file_chunk_data` feeds only a debug log's chunk count. So pulling the BM25 half into a single pass required no reordering of the other two, and `bm25_index` drops out of both helper signatures.

The one real constraint is that the batch has to run **before** `insert_chunks`, or it would delete the documents just written. That is called out in a comment at the call site.

## Tests

No new tests, because existing coverage already fails from both directions — I checked rather than assumed:

- Moving the batch delete after `insert_chunks` fails `update_modified_file` ("BM25 should find updated content") and `update_mixed_add_modify_delete` ("should find updated function name").
- Making `delete_by_files` a no-op fails `update_deleted_file` and `update_removes_old_chunks_when_modified_file_has_no_chunks`.

Adding more tests on top of that would be redundant.

772 `vera-core` tests, 92 `vera-cli`, `cargo fmt --check` clean, clippy unchanged at the pre-existing warnings.

## Note

`delete_by_chunk_id` has the same writer-per-call shape. It is not on a per-file loop today, so I have left it alone rather than widen this PR speculatively.

Fixes #82

---
## Summary by cubic
Batch-deletes BM25 documents for all changed files in one pass and runs the batch before any per-file mutations. Replaces per-file deletes and fixes a retry bug where dropping tombstones first could leave stale BM25 docs.

- `crates/vera-core/src/storage/bm25.rs`: adds `delete_by_files(&[&str])`; `delete_by_file` delegates. Tests cover multi-path deletion and the empty-slice no-op.
- `crates/vera-core/src/indexing/update.rs`: precomputes which prepared files need cleanup (modified or previously had chunks), collects those paths plus deleted ones, and issues one `delete_by_files` before any writes or `insert_chunks`. Removes BM25 deletion from `remove_file_chunk_data` and `remove_file_from_index`; these helpers drop the `Bm25Index` parameter. Avoids duplicate `get_chunks_by_file` reads.
- Perf (200-doc index): 10 files 216ms→32ms; 50 files 1.19s→29ms; 100 files 2.48s→37ms.

<sup>Written for commit 9142ae12b87c6073bbd3c78b4a5c60552c6d5ca0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/84?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup when files are deleted or modified, ensuring associated search entries and file data are removed consistently.
  - Prevented stale search results from remaining after file updates or deletions.
  - Improved handling of empty deletion requests without unnecessary processing.

- **Performance**
  - Streamlined bulk removal of search index entries for multiple files.
  - Reduced repeated processing during cleanup, improving efficiency when several files are changed or deleted.

### DIFF
--- a/crates/vera-core/src/indexing/update.rs
+++ b/crates/vera-core/src/indexing/update.rs
@@ -595,10 +595,10 @@ where
             Bm25Index::open(&bm25_dir).context("failed to open BM25 index for update")?;
 
         // Every path whose BM25 documents have to go is removed in one pass
-        // before any per-file mutation. `delete_by_file` allocates a writer, commits a segment and
-        // joins the merge threads, which costs tens of milliseconds per call
-        // however little is actually deleted. The batch must stay ahead of
-        // the `insert_chunks` below, or it would delete newly written docs.
+        // before any per-file mutation. `delete_by_file` allocates a writer,
+        // commits a segment and joins the merge threads, which costs tens of
+        // milliseconds per call however little is actually deleted. The batch
+        // must stay ahead of `insert_chunks`, or it would delete newly written docs.
         let mut bm25_deletions: Vec<&str> = deleted.iter().map(String::as_str).collect();
         let cleanup_chunk_data: Vec<bool> = prepared_files
             .iter()
@@ -642,7 +642,6 @@ where
                 .delete_file_state(&file.path)
                 .with_context(|| format!("failed to delete file state for {}", file.path))?;
         }
-
         for file in &prepared_files {
             if !file.references.is_empty() {
                 metadata_store
--- a/crates/vera-core/src/indexing/update.rs
+++ b/crates/vera-core/src/indexing/update.rs
@@ -797,17 +797,14 @@ fn remove_file_parse_data(metadata_store: &MetadataStore, file_path: &str) -> Re
     Ok(())
 }
 
-/// Remove chunk data (chunk metadata, vectors, BM25 entries) for a file.
+/// Remove vector and chunk metadata for a file.
+///
+/// The caller must delete the file's BM25 documents before calling this helper.
 fn remove_file_chunk_data(
     metadata_store: &MetadataStore,
     vector_store: &VectorStore,
     file_path: &str,
 ) -> Result<()> {
-    // Get chunk IDs for this file (needed for vector/BM25 deletion).
-    let chunks = metadata_store
-        .get_chunks_by_file(file_path)
-        .context("failed to get chunks for file deletion")?;
-
     // Delete from vector store using file prefix pattern.
     let prefix = format!("{file_path}:");
     vector_store
@@ -819,11 +816,7 @@ fn remove_file_chunk_data(
         .delete_chunks_by_file(file_path)
         .with_context(|| format!("failed to delete metadata for {file_path}"))?;
 
-    debug!(
-        file = %file_path,
-        chunks = chunks.len(),
-        "removed file chunk data from index"
-    );
+    debug!(file = %file_path, "removed file chunk data from index");
 
     Ok(())
 }
--- a/crates/vera-core/src/indexing/update.rs
+++ b/crates/vera-core/src/indexing/update.rs
@@ -594,25 +594,49 @@ where
         let bm25_index =
             Bm25Index::open(&bm25_dir).context("failed to open BM25 index for update")?;
 
-        // All parsing and embedding work is complete. Writes below publish the prepared update.
+        // Every path whose BM25 documents have to go is removed in one pass
+        // before any per-file mutation. `delete_by_file` allocates a writer, commits a segment and
+        // joins the merge threads, which costs tens of milliseconds per call
+        // however little is actually deleted. The batch must stay ahead of
+        // the `insert_chunks` below, or it would delete newly written docs.
+        let mut bm25_deletions: Vec<&str> = deleted.iter().map(String::as_str).collect();
+        let cleanup_chunk_data: Vec<bool> = prepared_files
+            .iter()
+            .map(|file| {
+                metadata_store
+                    .get_chunks_by_file(&file.path)
+                    .with_context(|| {
+                        format!("failed to inspect existing chunks for {}", file.path)
+                    })
+                    .map(|chunks| file.modified || !chunks.is_empty())
+            })
+            .collect::<Result<_>>()?;
+        bm25_deletions.extend(
+            prepared_files
+                .iter()
+                .zip(&cleanup_chunk_data)
+                .filter(|(_, cleanup)| **cleanup)
+                .map(|(file, _)| file.path.as_str()),
+        );
+
+        // All parsing, embedding, and read-only cleanup discovery is complete.
+        // Writes below publish the prepared update and must run to completion.
         cancellation.check()?;
+        bm25_index
+            .delete_by_files(&bm25_deletions)
+            .context("failed to delete BM25 entries for changed files")?;
+
         for file_path in &deleted {
-            remove_file_from_index(&metadata_store, &vector_store, &bm25_index, file_path)?;
+            remove_file_from_index(&metadata_store, &vector_store, file_path)?;
         }
 
         // A previous attempt may have failed after writing one store but before
         // committing the file hash. Cleaning every prepared path makes retries
         // idempotent for both modified and newly added files.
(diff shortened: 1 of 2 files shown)


citron07r:
Three findings from a local review pass. Two acted on, one declined with reasons.

**Tombstones dropped before the BM25 delete succeeded (valid, fixed in 9114c51).** This was a real regression I introduced and the best catch of the three. Moving the BM25 delete after the per-file loops meant it ran after `remove_file_from_index`, which for deleted files removes the file hash and index state — precisely what takes the path out of `tracked_files()`. If the batch then failed, the path would no longer classify as deleted on the next run, so its documents would linger with nothing left to trigger a retry. Before this PR the per-file `delete_by_file` ran *before* that removal, so the `?` left the tombstone intact and the next run retried; batching quietly inverted it.

Fixed by running the batch first, ahead of both loops. A failure there now leaves every store untouched, and it still sits ahead of `insert_chunks`, which is the other ordering constraint. Both are stated at the call site so the next person moving this code knows there are two directions to get wrong, not one.

**Direct tests for the batch API (valid, added).** Multi-path deletion taking all of each path's documents and none of anyone else's — `src/main.rs` has two documents in the fixture, so it also covers the several-docs-per-path case — plus the empty-slice no-op, which the update path depends on since it now calls this unconditionally. I had argued the existing integration coverage was enough; it does catch the integration, but it does not pin the new API's own contract, so this was a fair ask.

**Typed `thiserror` error for the new API (declined).** The guidance is right in general, but `crates/vera-core/src/storage/` has **zero** uses of `thiserror` — `bm25.rs` imports `anyhow::{Context, Result}` and all ten of its public methods return `anyhow::Result`, as does the rest of the module. `delete_by_files` is an extraction of `delete_by_file`'s body and returns exactly what it returned.

Converting one new function would make it the only typed-error surface in the module and force callers to handle two error styles from the same type. That is a module-wide refactor with a real design decision behind it — the `{0:#}` cause-chain trap from #42 is the kind of thing it would have to get right — so it belongs in its own change with maintainer buy-in, not as a side effect of a perf PR. Happy to open an issue for it if that is wanted.

Re-verified after the reordering that coverage still fails from both directions: no-op'ing `delete_by_files` fails `update_deleted_file`, `update_removes_old_chunks_when_modified_file_has_no_chunks` and the new batch test; reversing it against `insert_chunks` fails `update_modified_file` and `update_mixed_add_modify_delete`.

774 `vera-core` tests, 92 `vera-cli`, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

citron07r:
Two findings since the last round. One was a real vacuity I introduced with @lemon07r's help, fixed in 0db0a8c; the other does not match the current ordering.

**The deleted-path assertion was vacuous (valid, fixed).** 88827cd switched the query from `"fn"` to `"hello"` to make it cover the deleted paths, but `"hello"` occurs *only* in the two documents being deleted (`src/main.rs:0` "Hello, world!" and `src/lib.py:0` "hello_world" / "Hello from Python"). After the deletion the search returns nothing, so `hits.iter().all(...)` iterated an empty collection and verified nothing.

Measured after deletion, to be sure rather than to reason about it:

```
query "hello" -> 0 hits
query "name"  -> 1 hit
```

`"name"` appears in `src/main.rs:1` (deleted) and `fuzz/Cargo.toml:0` (kept), so it matches a survivor while still being able to surface a deleted chunk if one leaked through. There is now an explicit non-empty assertion in front of the loop, so a future query change cannot quietly make it vacuous again.

**"Cleanup fails before this call, leaving BM25 stale" (does not apply to the current code).** That sequence describes the ordering in 018e0ff, where the batch ran *after* the per-file cleanups. 9114c51 moved it ahead of both loops precisely because of the tombstone problem, so at head there is no cleanup before the batch call and no `?` that can skip it.

The inverse is worth stating, since it is the failure this ordering does have: if a per-file cleanup fails partway, BM25 documents are already gone while some chunks and vectors remain. That is the better of the two, and deliberately so. The file hash is removed last, so the path stays in `tracked_files()`, the next run classifies it as changed again, and the BM25 entries are re-inserted. Self-healing. The other ordering loses the tombstone first and leaves orphan BM25 documents with nothing left to trigger a retry.

If the concern is the partially-cleaned intermediate state rather than the recovery path, that is a fair thing to want, but it needs the three stores under one transaction, which they cannot be: tantivy and the two SQLite databases have no shared transaction. Worth its own issue if you want it tracked.

774 `vera-core` tests, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

PR #86, opened 2026-08-19
perf(storage): resolve KNN rowids in one query instead of one per hit

### LINKED ISSUE #85: Vector search does one uncached query per KNN hit, and its interpolated LIMIT thrashes the statement cache
## Summary

Two issues in `VectorStore::search` (`crates/vera-core/src/storage/vector.rs`), both on the hot search path.

## 1. One uncached query per KNN hit

After the KNN query returns, each result's rowid is mapped back to a `chunk_id` with a separate `Connection::query_row` inside the result loop. `query_row` on a `Connection` re-prepares the statement every call, and the loop runs once per candidate — up to `MAX_KNN_K` (4096) on a filtered natural-language query.

Measured against this repository's own `vectors.db`, comparing per-row lookups with a single `IN (...)` batch:

```
n=  100   per-row  0.19 ms   batch 0.07 ms   2.8x
n= 1000   per-row  1.63 ms   batch 0.44 ms   3.7x
n= 4096   per-row  6.64 ms   batch 1.86 ms   3.6x
```

Roughly 4.8 ms saved per search at the ceiling. Note these numbers were taken through Python's sqlite3, which caches prepared statements internally — so they *understate* the real gap, since the production path re-prepares on every call.

The safe shape is to leave the vec0 KNN query completely untouched and do one `SELECT rowid, chunk_id FROM chunk_id_map WHERE rowid IN (...)` afterwards, then map. Folding `chunk_id_map` into the vec0 query as a JOIN would risk losing the KNN optimization and should be avoided.

## 2. `LIMIT` is interpolated into a `prepare_cached` statement

```
.prepare_cached(&format!(
    "SELECT v.rowid, v.distance
     FROM vec_chunks v
     WHERE v.embedding MATCH ?1
     ORDER BY v.distance
     LIMIT {limit}"
))
```

The SQL text varies with `limit`, so every distinct limit is a distinct cache key. rusqlite's statement cache is an LRU with a default capacity of **16** (`STATEMENT_CACHE_DEFAULT_CAPACITY`, `rusqlite-0.39.0/src/lib.rs:160`), and Vera never calls `set_prepared_statement_cache_capacity`. Since `candidate_pool` derives the limit from the caller's requested limit, varied searches produce varied SQL, so this both misses the cache and evicts the *other* cached statements on that connection — including the ones that do have stable text.

Either bind the limit as a parameter, or use plain `prepare` so it stops occupying cache slots it can never reuse. Worth confirming the sqlite-vec version in use accepts a bound `LIMIT` in a KNN query before choosing binding — if it does not, `prepare` is the fix.

## Priority

Neither is dramatic on its own. Item 1 is the larger of the two and is bounded by a known ceiling; item 2 is small but nearly free to fix and has a side effect on unrelated queries sharing the connection. Filing them together since both live in the same function and would be one PR.

Related: #78 and #79 covered the metadata-side and delete-side query plans; this is the remaining search-side item from the same review.

citron07r:
## Problem

Two things in `VectorStore::search`, both on the hot search path.

**One uncached query per KNN hit.** After the KNN query returns, each hit's rowid was mapped back to a `chunk_id` with a separate `Connection::query_row` inside the result loop. `query_row` on a `Connection` re-prepares the statement every call, and the loop runs once per candidate — up to `MAX_KNN_K` (4096) on a filtered natural-language query.

**`LIMIT` interpolated into a `prepare_cached` statement.** The SQL text varied with `limit`, so every distinct limit was a distinct cache key. rusqlite's statement cache is an LRU with a default capacity of 16 (`STATEMENT_CACHE_DEFAULT_CAPACITY`), and Vera never raises it — so these entries could never hit, and they evicted the statements that *do* have stable text on the same connection.

## Change

One batched `SELECT rowid, chunk_id FROM chunk_id_map WHERE rowid IN (...)` after the KNN query, then re-project.

The vec0 KNN query itself is deliberately left alone. Joining `chunk_id_map` into it would risk losing the KNN optimization, which would cost far more than this saves.

Both statements switch from `prepare_cached` to `prepare`, since both have text that varies per call. sqlite-vec does not accept a bound parameter for the KNN `LIMIT`, so the interpolation stays — it just stops occupying a cache slot it can never reuse.

## Measurement

Against this repository's own `vectors.db`, per-row lookups versus one batch:

| Hits | Per-row | Batched |
|---|---|---|
| 100 | 0.19 ms | 0.07 ms |
| 1000 | 1.63 ms | 0.44 ms |
| 4096 | 6.64 ms | 1.86 ms |

~4.8 ms per search at the ceiling. Worth being explicit that these were taken through Python's sqlite3, which caches prepared statements internally — so they **understate** the real gap, because the production path re-prepared on every call.

End to end on a real index, search output is byte-identical to the released binary.

## The part that needed a real test

Batching means the mapping comes back as a `HashMap`, so the results have to be re-projected into the KNN distance order. That is the same shape as the bug in #73, so it gets the same treatment.

`search_pairs_each_chunk_id_with_its_own_distance` inserts three vectors so that rowid order is the exact reverse of distance order, then asserts the chunk_id/distance **pairing** — not merely that distances ascend, which holds for any implementation emitting one result per hit in hit order.

Verified it catches the regression rather than assuming: zipping the hits against SQL order returns `["far", "mid", "near"]` instead of `["near", "mid", "far"]`, and the test fails. The pre-existing `nearest_neighbor_self_query` passes either way, which is why the new one was needed.

773 `vera-core` tests, 97 `vera-cli`, `cargo fmt --check` clean, clippy unchanged at the pre-existing warnings.

Fixes #85

---
## Summary by cubic
Resolve KNN rowids in one batched query and stop caching SQL that varies per call. Old behavior: one uncached row lookup per hit and `prepare_cached` with interpolated `LIMIT`; new behavior: a single `IN (...)` lookup and `prepare`, cutting lookup time from 6.64 ms to 1.86 ms at 4096 hits with identical results.

- Keep the KNN query unchanged; do not join `chunk_id_map` to preserve `sqlite-vec` optimization.
- Interpolate `LIMIT` (cannot bind with `sqlite-vec`) and use `prepare` to avoid polluting `rusqlite`’s 16-entry cache.
- Re-project results to distance order; tests assert chunk_id/distance pairing and strictly increasing distances.
- Batch lookup handles 4096 bound parameters; search returns an error if a rowid lacks a `chunk_id` mapping.
- Refactor: `chunk_ids_for_rowids` now takes `&[i64]`; the caller projects rowids once (bounded by MAX_KNN_K).
- No API changes or migrations. Output matches the released binary on a real index.

<sup>Written for commit e003eff50f2970c63f70559e9b9998426e137e85. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/86?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured vector search results retain the correct chunk identifiers and proximity distances.
  * Preserved nearest-result ordering in search results.
  * Improved handling of searches with dynamically adjusted result limits.
  * Added regression coverage for missing mappings and parameter limits.

### DIFF
--- a/crates/vera-core/src/storage/vector.rs
+++ b/crates/vera-core/src/storage/vector.rs
@@ -554,9 +554,15 @@ mod tests {
         // reverse of distance order, then assert the pairing — not just that
         // distances ascend, which holds for any implementation that emits one
         // result per hit in hit order.
+        // Distances must be strictly increasing, or the KNN order between two
+        // equidistant vectors is unspecified and the assertion below would be
+        // decided by tie-breaking rather than by distance.
+        //   near 0.0   mid ~0.894   far ~1.414   (L2 from the query)
+        // Inserted in reverse of that order, so rowid order is not distance
+        // order and a SQL-ordered result cannot pass by accident.
         let store = VectorStore::open_in_memory(4).unwrap();
         store.insert("far", &[0.0, 0.0, 1.0, 0.0]).unwrap();
-        store.insert("mid", &[0.0, 1.0, 0.0, 0.0]).unwrap();
+        store.insert("mid", &[0.6, 0.8, 0.0, 0.0]).unwrap();
         store.insert("near", &[1.0, 0.0, 0.0, 0.0]).unwrap();
 
         let results = store.search(&[1.0, 0.0, 0.0, 0.0], 3).unwrap();
@@ -575,8 +581,8 @@ mod tests {
         );
         for pair in results.windows(2) {
             assert!(
-                pair[0].distance <= pair[1].distance,
-                "distances must ascend: {:?}",
+                pair[0].distance < pair[1].distance,
+                "distances must strictly ascend, or the order is a tie-break: {:?}",
                 results.iter().map(|r| r.distance).collect::<Vec<_>>()
             );
         }
--- a/crates/vera-core/src/storage/vector.rs
+++ b/crates/vera-core/src/storage/vector.rs
@@ -262,7 +262,8 @@ impl VectorStore {
         // Resolve the rowids in one query instead of one per hit. The KNN
         // query above is deliberately left alone: joining `chunk_id_map` into
         // it would risk losing the vec0 KNN optimization.
-        let chunk_ids = self.chunk_ids_for_rowids(&hits)?;
+        let rowids: Vec<i64> = hits.iter().map(|(rowid, _)| *rowid).collect();
+        let chunk_ids = self.chunk_ids_for_rowids(&rowids)?;
 
         hits.iter()
             .map(|(rowid, distance)| {
@@ -279,12 +280,12 @@ impl VectorStore {
     }
 
     /// Resolve many rowids to their chunk ids in a single query.
-    fn chunk_ids_for_rowids(&self, hits: &[(i64, f64)]) -> Result<HashMap<i64, String>> {
-        if hits.is_empty() {
+    fn chunk_ids_for_rowids(&self, rowids: &[i64]) -> Result<HashMap<i64, String>> {
+        if rowids.is_empty() {
             return Ok(HashMap::new());
         }
 
-        let placeholders = std::iter::repeat_n("?", hits.len())
+        let placeholders = std::iter::repeat_n("?", rowids.len())
             .collect::<Vec<_>>()
             .join(",");
         // Text varies with the number of ids, so plain `prepare` here too.
@@ -296,10 +297,9 @@ impl VectorStore {
             .context("failed to prepare chunk id lookup")?;
 
         let mapped = stmt
-            .query_map(
-                rusqlite::params_from_iter(hits.iter().map(|(rowid, _)| *rowid)),
-                |row| Ok((row.get::<_, i64>(0)?, row.get::<_, String>(1)?)),
-            )
+            .query_map(rusqlite::params_from_iter(rowids.iter()), |row| {
+                Ok((row.get::<_, i64>(0)?, row.get::<_, String>(1)?))
+            })
             .context("failed to query chunk ids")?
             .collect::<std::result::Result<HashMap<_, _>, _>>()
             .context("failed to collect chunk ids")?;
@@ -601,8 +601,8 @@ mod tests {
                 .unwrap();
         }
 
-        let hits: Vec<(i64, f64)> = (1..=4096).map(|rowid| (rowid, 0.0)).collect();
-        let mapped = store.chunk_ids_for_rowids(&hits).unwrap();
+        let rowids: Vec<i64> = (1..=4096).collect();
+        let mapped = store.chunk_ids_for_rowids(&rowids).unwrap();
 
         assert_eq!(mapped.len(), 4096);
         assert_eq!(mapped.get(&1).map(String::as_str), Some("chunk-1"));
--- a/crates/vera-core/src/storage/vector.rs
+++ b/crates/vera-core/src/storage/vector.rs
@@ -589,6 +589,46 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


citron07r:
Four findings. One was a real defect in my test and is fixed in dfc536e; the other three I am declining, with reasons.

**Equidistant fixture made the ordering assertion a tie-break (valid, fixed).** This was the important one and it was my mistake. `mid` `[0,1,0,0]` and `far` `[0,0,1,0]` are both exactly √2 from the query `[1,0,0,0]`, so which came back first was decided by tie-breaking rather than by distance. Worse, my earlier check that the test catches a SQL-ordered result was resting partly on that tie instead of on the property being tested.

`mid` is now `[0.6,0.8,0,0]`, giving 0.0 / ~0.894 / ~1.414, and the monotonicity assertion requires a *strict* increase so a future fixture cannot reintroduce a tie unnoticed. Re-ran the regression check with unambiguous distances: the test still fails on a SQL-order zip, `["far","mid","near"]` against `["near","mid","far"]`.

**Typed `thiserror` error for the new helper (declined).** Same answer as on #84, and the same evidence: `crates/vera-core/src/storage/` contains **zero** uses of `thiserror` — on master as well as on this branch. `vector.rs` imports `anyhow::{Context, Result}` and every public method returns `anyhow::Result`. `chunk_ids_for_rowids` is an extraction of code that already returned that.

Introducing one typed error here would make it the only typed-error surface in the module and force callers to handle two error styles from the same type. Converting the module is a real change with a decision behind it — the `{0:#}` cause-chain trap from #42 is exactly what it would have to get right — so it belongs in its own PR with maintainer buy-in, not as a side effect of a batching change. Happy to open an issue for it.

**Cap the `IN` list to the host-parameter limit (declined, with a caveat worth recording).** The default bound holds with room: `hits.len()` is capped by `MAX_KNN_K = 4096`, and the bundled SQLite 3.51.3 defaults `SQLITE_MAX_VARIABLE_NUMBER` to 32766 — 8x headroom.

The caveat is real though, and it is the one thing in this thread I had not considered: `libsqlite3-sys`'s build script honours a `SQLITE_MAX_VARIABLE_NUMBER` environment variable, so a packager could compile a lower limit. If that is a configuration the project wants to support, the fix belongs in one place rather than here — `MetadataStore::get_chunks_by_ids` (in #73) binds up to the same 4096 from the same KNN bound, so guarding only this call site would leave the sibling exposed and the codebase inconsistent about it. I would rather raise it as its own issue covering both than half-fix it here. Say the word and I will file it.

**The minor at line 291** arrived with only its category header and no finding text, so there is nothing for me to evaluate. Happy to look if it gets re-posted.

773 `vera-core` tests, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

citron07r:
Valid, fixed in e003eff. Rebased onto cfd2aa9 rather than over it.

`chunk_ids_for_rowids` took `&[(i64, f64)]` but only ever read the rowid, so the signature overstated what it needed, and the boundary test added in cfd2aa9 had to invent `0.0` distances to call it. It now takes `&[i64]`, and that test builds `(1..=4096).collect()` directly.

The caller projects the rowids once before the call. That is one allocation bounded by `MAX_KNN_K`, which is a fair trade for a helper whose signature states its actual contract.

796 `vera-core` tests, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

PR #87, opened 2026-08-19
fix(parsing): select the signature grammar by path, like indexing does

### LINKED ISSUE #75: extract_signature picks the grammar by language, not by path (tsx split)
PR #58 made parsing path-aware: `.tsx` files now parse with the tsx grammar via `tree_sitter_grammar_for_path`. Signature extraction (`extract_signature` in the parsing extractor) still selects the grammar from the `Language` enum alone, so for `.tsx` files it re-parses (or scans) with the plain TypeScript grammar.

Impact is probably cosmetic (signature strings on chunks), but the two paths can disagree on files with JSX-heavy syntax.

Verify whether signature extraction actually diverges on real .tsx files; if it does, route it through the same path-aware grammar selection.

citron07r:
## Verified

The premise holds. `extract_signature_inner` called `tree_sitter_grammar(lang)`, selecting by the `Language` enum alone, while indexing goes through `tree_sitter_grammar_for_path`. For a `.tsx` file the two parsed the same content with different grammars.

## Does it actually diverge?

Yes, but not where I first expected, and the honest answer about impact is more limited than the issue assumes.

`extract_signature` finds the byte offset of the first body node and returns everything before it. JSX lives *inside* bodies, so the grammars agree on where a body starts. I tried nine realistic samples — components, generic-vs-JSX ambiguity (`<T,>`), arrow components, default parameters containing JSX, class property initializers, type assertions, fragments — and every one produced an identical signature under both grammars.

The divergence appears when JSX precedes a declaration in the same content. Then the TypeScript grammar fails to locate a body within the depth limit at all, and `extract_signature` silently falls back to dumping raw lines:

```
typescript: "const A = <p>{ q }</p>;\nfunction Later(x: number) {\n  return x;\n[... 1 more lines]"
tsx:        "const A = <p>{ q }</p>;\nfunction Later(x: number) { ... }"
```

**I could not reach that through the normal chunking path.** The chunker gives each top-level declaration its own chunk, so a chunk's content does not start with JSX ahead of a declaration. Indexing the case above produces four chunks, and the `Toolbar` chunk begins at the declaration — `vera search --compact` output is identical before and after this change.

So: a real inconsistency in a public API, with a demonstrable output difference at the function level, but no user-visible symptom I could produce. I would rather say that plainly than dress it up.

## Change

`extract_signature` takes a `file_path` and routes through `tree_sitter_grammar_for_path`, the same selection indexing uses. All four call sites already had a path to hand (`SearchResult::file_path`, `Chunk::file_path`), so nothing needed plumbing.

The regression test pins both directions: the same content under a `.tsx` path yields a stripped signature, and under a `.ts` path keeps the old fallback — which is correct there, since outside a `.tsx` file the angle brackets really are type syntax. Confirmed the test fails when the grammar selection is reverted.

774 `vera-core` tests, 97 `vera-cli`, `cargo fmt --check` clean, clippy unchanged at the pre-existing warnings.

Fixes #75

---
## Summary by cubic
Selects the signature grammar by file path to match indexing. Previously grammar came from `Language` only; `.tsx` could parse as TypeScript and fall back to raw lines when JSX precedes a declaration. Now `.tsx` uses the TSX grammar, producing stripped signatures and aligning outputs.

- Adds `extract_signature_for_path(content, lang, file_path)` and routes `vera-core` (including `retrieval/type_relations`), `vera-cli`, and `vera-mcp` to it; keeps `extract_signature` for callers without a path.
- Behavior: function-level output now matches indexing for `.tsx`; chunked results remain unchanged. Tests add a `.tsx` vs `.ts` regression and pin the exact `.ts` first-three-lines fallback.

<sup>Written for commit d62119e1c759bd5d42af8d8670ca5e4e487048bd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/87?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved code signature extraction for files with extensions such as `.tsx`.
  * Preserved reliable fallback behavior for TypeScript files.
  * Updated result displays and relationship information to use file-specific parsing for more accurate output.

### DIFF
--- a/crates/vera-cli/src/helpers.rs
+++ b/crates/vera-cli/src/helpers.rs
@@ -341,7 +341,7 @@ pub fn output_results(
     let compacted: Vec<String> = if compact {
         results
             .iter()
-            .map(|r| extract_signature(&r.content, r.language))
+            .map(|r| extract_signature(&r.content, r.language, &r.file_path))
             .collect()
     } else {
         Vec::new()
--- a/crates/vera-cli/src/helpers.rs
+++ b/crates/vera-cli/src/helpers.rs
@@ -335,13 +335,13 @@ pub fn output_results(
     compact: bool,
     budget: usize,
 ) {
-    use vera_core::parsing::signatures::extract_signature;
+    use vera_core::parsing::signatures::extract_signature_for_path;
 
     // When compact mode is on, pre-compute signature-only content for each result.
     let compacted: Vec<String> = if compact {
         results
             .iter()
-            .map(|r| extract_signature(&r.content, r.language, &r.file_path))
+            .map(|r| extract_signature_for_path(&r.content, r.language, &r.file_path))
             .collect()
     } else {
         Vec::new()
--- a/crates/vera-core/src/parsing/signatures.rs
+++ b/crates/vera-core/src/parsing/signatures.rs
@@ -185,12 +185,13 @@ mod tests {
         );
 
         // The same content under a .ts path keeps the old behaviour, which is
-        // correct: there the angle brackets really are type syntax.
+        // correct: there the angle brackets really are type syntax. Assert the
+        // exact `first_n_lines` output rather than merely "not the tsx one" —
+        // a differently malformed signature would satisfy a negative check.
         let ts = extract_signature(code, Language::TypeScript, "Widget.ts");
-        assert_ne!(ts, tsx);
-        assert!(
-            !ts.contains("{ ... }"),
-            "the .ts path should fall back, got: {ts}"
+        assert_eq!(
+            ts,
+            "const A = <p>{ q }</p>;\nfunction Later(x: number) {\n  return x;\n[... 1 more lines]"
         );
     }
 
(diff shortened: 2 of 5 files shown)


citron07r:
Valid, fixed in 111bd6f.

The `.ts` branch only required the output to differ from the tsx one and to omit `"{ ... }"`. A differently malformed signature would have satisfied both, which makes it a check that the two paths disagree rather than a check that the fallback is correct.

It now asserts the full `first_n_lines` result:

```
const A = <p>{ q }</p>;
function Later(x: number) {
  return x;
[... 1 more lines]
```

That is the documented fallback behaviour, so the test now pins what the `.ts` path is supposed to produce instead of merely what it is not.

773 vera-core tests pass, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

PR #88, opened 2026-08-19
docs(parsing): pin the naming convention for dotted JSX components

### LINKED ISSUE #76: JSX member-expression components index under the rightmost segment only (<Foo.Bar/> becomes Bar)
Follow-up from #58. JSX call sites for dotted component names (`<Foo.Bar />`) are stored under the rightmost segment (`Bar`), matching how the extractor treats member expressions elsewhere. `<foo.Bar />` (lowercase root) is also kept as `Bar`, which assumes any capitalized member is a component.

Decide and document the convention: either keep rightmost-segment naming (current), or store the full member path (`Foo.Bar`) so references/definitions line up with how developers actually grep for dotted components. Whichever way, the behavior should be documented in the parsing docs and covered by a test that pins the choice.

citron07r:
## Verified

Both observations in the issue are accurate. `extract_callee` returns the rightmost identifier, so `<Icons.Arrow />` records `Arrow`. And `is_jsx_host_element` returns early for a `member_expression`, so `<icons.Arrow />` with a lowercase root is kept as a component and also recorded as `Arrow`.

The issue asks for a decision, documentation and a pinning test rather than reporting a defect. This keeps the current behaviour, and writes down why.

## The decision: rightmost segment

The deciding argument is linkage, not developer grep habits.

Definitions are indexed under their bare declared name. A component reached as `Icons.Arrow` is declared somewhere as `export function Arrow`, and `find_callers` matches `lower(callee)` against that name. Storing the full member path would leave the reference matching **no** definition — reintroducing precisely the invisibility that indexing JSX call sites was added to fix in #58.

It also keeps JSX consistent with the rest of the extractor, where `obj.method()` already records `method`. Special-casing JSX to store dotted paths would make `vera references` behave differently depending on whether the call site happened to be JSX.

Verified end-to-end on a real index:

```
$ vera references Arrow
src/toolbar.tsx:4-7 function:Toolbar
    return <div><Icons.Arrow /><span /></div>;
```

## Lowercase roots

`<icons.Arrow />` staying a component is correct rather than an oversight. TypeScript's intrinsic test applies only to identifiers — a property access is never intrinsic, whatever the case of its root. So a dotted tag is always a value lookup, and treating `icons.Arrow` as an HTML tag would drop a real reference.

## The trade-off, stated rather than hidden

Two components sharing a final segment (`Icons.Arrow` and `Shapes.Arrow`) collapse onto one name. Separating them needs import and module resolution, which this layer does not do — it reads one file at a time with no cross-file symbol table. That limitation is now in the module documentation rather than being something a reader has to infer from the code.

## Test

`jsx_dotted_components_index_under_the_rightmost_segment` pins all four properties: the uppercase-rooted dotted component indexes as its rightmost segment, the lowercase-rooted one does too, no callee carries a dotted path, and intrinsic host elements (`div`, `span`) stay out of the call graph. If someone later switches to full member paths, that test is where the decision gets re-opened deliberately instead of silently.

773 `vera-core` tests, `cargo fmt --check` clean, clippy unchanged.

Fixes #76

---
## Summary by cubic
Pins and documents the naming convention for dotted JSX components in `vera-core` parsing to preserve reference → definition links. Behavior is unchanged: we index the rightmost segment (e.g., `<Icons.Arrow />` → `Arrow`), consistent with `obj.method()`.

- Documents in `references.rs` that dotted JSX tags index by their rightmost segment; member expressions with lowercase roots remain components; bare host elements are excluded by the lowercase or hyphen rule (e.g., `div`, `my-element`, `My-element`).
- Adds test `jsx_dotted_components_index_under_the_rightmost_segment` asserting the exact callee set (`Arrow`, `Chevron`), including `<My-element />` to cover the hyphen rule; ensures no dotted paths or host elements leak into references.
- Notes the trade-off: components sharing a final segment collapse onto one name; import/module resolution is out of scope for this layer.

<sup>Written for commit bff96b344b0c09f6d86fbf8b96b355d4aa649f67. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/88?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Added documentation clarifying how dotted callees and JSX member expressions are normalized, including known symbol-resolution limitations.

* **Tests**
  * Added TSX coverage for dotted JSX components.
  * Verified handling of uppercase and lowercase member expressions, while excluding intrinsic elements and full dotted paths from call-site references.

### DIFF
--- a/crates/vera-core/src/parsing/references.rs
+++ b/crates/vera-core/src/parsing/references.rs
@@ -3,6 +3,29 @@
 //! Walks the AST looking for function/method call expressions and records
 //! the callee name, source file, and line number. These lightweight edges
 //! power `vera references`, `vera impact`, and `vera dead-code`.
+//!
+//! # Naming convention for dotted callees
+//!
+//! A callee is stored under its **rightmost segment**. `obj.method()` records
+//! `method`, and JSX follows the same rule, so `<Icons.Arrow />` records
+//! `Arrow` rather than `Icons.Arrow`.
+//!
+//! This is what makes a reference link to a definition. Definitions are
+//! indexed under their bare declared name — a component reached as
+//! `Icons.Arrow` is declared somewhere as `export function Arrow` — and
+//! `find_callers` matches the callee against that name. Storing the full
+//! member path would leave the reference matching no definition, which is the
+//! invisibility that indexing JSX call sites exists to fix.
+//!
+//! A lowercase root does not change this: `<icons.Arrow />` is a member
+//! expression, and a member expression is always a value lookup. Only a bare
+//! lowercase identifier is an intrinsic host element (see
+//! `is_jsx_host_element`).
+//!
+//! The trade-off is deliberate: two components sharing a final segment
+//! (`Icons.Arrow` and `Shapes.Arrow`) collapse onto one name. Separating them
+//! needs import and module resolution, which this layer does not do — it reads
+//! one file at a time with no cross-file symbol table.
 
 use crate::types::Language;
 
--- a/crates/vera-core/src/parsing/references.rs
+++ b/crates/vera-core/src/parsing/references.rs
@@ -18,9 +18,14 @@
 //! invisibility that indexing JSX call sites exists to fix.
 //!
 //! A lowercase root does not change this: `<icons.Arrow />` is a member
-//! expression, and a member expression is always a value lookup. Only a bare
-//! lowercase identifier is an intrinsic host element (see
-//! `is_jsx_host_element`).
+//! expression, and a member expression is always a value lookup — the
+//! intrinsic test never applies to one.
+//!
+//! For *bare* tags there are two intrinsic forms, not one: a name starting
+//! with a lowercase ASCII letter (`<div />`), and any name containing a hyphen
+//! whatever its case (`<My-element />`, `<my-element />`), which is the
+//! custom-element form. Both are host elements and stay out of the call graph.
+//! See `is_jsx_host_element`.
 //!
 //! The trade-off is deliberate: two components sharing a final segment
 //! (`Icons.Arrow` and `Shapes.Arrow`) collapse onto one name. Separating them
--- a/crates/vera-core/src/parsing/tests.rs
+++ b/crates/vera-core/src/parsing/tests.rs
@@ -397,6 +397,67 @@ export function App() {
     );
 }
 
+#[test]
+fn jsx_dotted_components_index_under_the_rightmost_segment() {
+    // Pins the naming convention for dotted JSX components.
+    //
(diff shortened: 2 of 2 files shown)


citron07r:
Both findings valid, fixed in f0d46b6. cubic and CodeRabbit caught the same doc gap independently.

**The hyphen rule was missing from the documentation.** The paragraph said only a bare lowercase identifier is an intrinsic host element, which is half of what `is_jsx_host_element` implements. Any bare name containing a hyphen is intrinsic whatever its case, because that is the custom-element form and TypeScript emits `<My-element />` as a string rather than a value lookup. Documenting one condition and not the other would have implied every non-dotted capitalized tag is a component, which is the opposite of what the code does.

The paragraph now states both intrinsic forms for bare tags, and keeps them separate from the member-expression case, where the intrinsic test never applies at all.

**The test used `contains`, so extra references would have passed.** Correct, and this is the failure mode most worth catching here: a root segment leaking in as its own reference (`Icons`, `icons`) is exactly what a dotted-name change would produce, and a containment check cannot see it. The test now asserts the exact set `["Arrow", "Chevron"]`.

The fixture also gained `<My-element />` so the hyphen rule is covered by a test rather than only described in prose. It sits alongside `<div>` and `<span>`, so the assertion now exercises both intrinsic forms.

773 vera-core tests pass, `cargo fmt --check` clean, clippy unchanged.

citron07r:
Correct, and thank you for catching it. Fixed in 9633997.

The fixture did not contain `<My-element />`. The edit that was meant to add it never reached the file, and I did not check before writing the comment that says it covers the hyphen rule. So the claim was in three places (the code comment, the commit message for f0d46b6, and my reply above saying "the fixture also gained `<My-element />`") and true in none of them. The exact-set assertion passed for an unrelated reason.

The tag is in the fixture now, and the assertion demonstrably depends on it. Removing `|| text.contains('-')` from `is_jsx_host_element`:

```
left:  ["Arrow", "Chevron", "My-element"]
right: ["Arrow", "Chevron"]
```

So a hyphenated tag leaking into the call graph now fails the test rather than being described in prose.

Worth noting this is the second time in this PR that the gap was between what the test claimed and what it checked, which is what the exact-set change was supposed to fix in the first place. Verifying by breaking the production rule, rather than by reading the test, is what caught it both times.

773 `vera-core` tests, `cargo fmt --check` clean, clippy unchanged at the 5 pre-existing warnings.

Issue #117, opened 2026-08-20
MCP and serve diverge from the CLI: no staleness warning, per-cycle model rebuild, per-request model load


citron07r:
Companion to #116, from the same audit. Three divergences where the MCP surface behaves worse than the CLI, all pre-existing.

## 1. MCP never tells the agent the index is stale (high)

`warn_if_index_stale` / `detect_staleness` appear only in `vera-cli` (`helpers.rs:67`, called at `helpers.rs:186` and `overview.rs:20`). Grep finds **no caller in `vera-mcp`**.

`ensure_index_and_watcher` only indexes when the directory is missing:

````rust
if !index_dir.exists() { ... index_repository(...) ... }
````

Scenario: a repo was indexed once, then `git pull` brings 500 commits. `vera mcp` starts, `search_code` sees `.vera/` exists, skips indexing, and searches months-old content. The watcher it then starts only observes changes from that moment on, so nothing ever backfills. The agent receives stale snippets with `isError: false` and presents them as current repository facts.

`vera search` on the same repo prints:

````
warning: index may be stale: N added, M modified. Search and grep only cover indexed files.
````

This is the divergence hardest for a user to notice, because both surfaces return plausible results and only one of them is honest about their age.

## 2. The MCP auto-watcher rebuilds the embedding model on every debounce cycle (high)

`vera-mcp/src/watcher.rs`: the recursive watch covers the whole tree with `.vera` as the only filter, and `run_update_blocking` creates a **new tokio runtime and a new ONNX session** every cycle (`create_dynamic_provider` has no cache).

Scenario: an agent runs `search_code` once, which starts the watcher. The user then runs `cargo build`. `target/` churns, every debounce window sees changes that are not under `.vera`, and each one spawns a thread, builds a runtime, builds an ONNX session, walks discovery, finds nothing indexable, and tears it all down. This repeats for the whole build and every build after it, visible only at `debug` level.

On Linux the recursive watch also registers inotify watches for `target/` and `node_modules/`. If `max_user_watches` is hit, `start_watching` fails and the caller only `warn!`s, so the index silently stops updating for the rest of the session.

Suggested: exclude the same set `discovery` already excludes rather than only `.vera`, and hoist the runtime and provider out of the per-cycle closure.

## 3. `vera serve` reloads the ONNX model on every request by default (high)

`--idle-timeout` defaults to `"0"`, and `0` maps to `None`, which makes both acquire paths bypass the cache and call `create_dynamic_provider` per request.

[...]

PR #130, opened 2026-08-20
fix(mcp): surface a stale-index warning in tool results

citron07r:
Part of #117.

## The bug

`warn_if_index_stale` / `detect_staleness` exist only in `vera-cli`
(`crates/vera-cli/src/helpers.rs:67`, called from `helpers.rs:183` and
`commands/overview.rs:20`). There is no caller anywhere in `vera-mcp`.

`ensure_index_and_watcher` (`crates/vera-mcp/src/tools.rs:632` at the base commit)
indexes only when the index directory is missing. After a `git pull`, `.vera/`
exists, so `search_code` skips indexing and searches stale content. The watcher it
starts observes changes from that moment forward, so nothing backfills what was
already missed.

The agent gets stale snippets with `isError: false` and no indication that they are
stale, and presents them as current repository facts. `vera search` on the same
repo, at the same moment, prints a warning.

## The fix

The notice is attached to the tool result itself, as a **trailing content block**.

Why not `tracing::warn!`: MCP has no out-of-band channel the client reads, so a
log line never reaches the agent. It is the agent that needs to know.

Why a separate trailing block rather than a line prepended to the result text:

- `content[0].text` stays byte-identical to the serialized results, so a client
  that parses it as JSON is unaffected. Prepending would break it; wrapping the
  array in an object would change the shape for every existing consumer.
- The notice is not part of the results string, so it is outside
  `MCP_OUTPUT_BUDGET` (`tools.rs:31`) and cannot silently push results out. The
  budget still gets its full 20,000 characters.
- Trailing rather than leading keeps `content[0]` addressable while still placing
  the warning at the end of the tool result, where a model reads it.

The wording is `IndexFreshness::stale_warning`
(`crates/vera-core/src/indexing/freshness.rs:55`), which the CLI's
`warn_if_index_stale` now also calls, so the two surfaces cannot drift:

    warning: index may be stale: 6 added, 59 modified. Search and grep only cover indexed files. Run `vera update .` or `vera watch .`.

The CLI's output is unchanged, checked against the released binary on the same
repository at the same moment:

    $ VERA_NO_UPDATE_CHECK=1 vera overview 2>&1 >/dev/null | head -1        # vera 1.0.0
    warning: index may be stale: 4 added, 33 modified. Search and grep only cover indexed files. Run `vera update .` or `vera watch .`.
    $ VERA_NO_UPDATE_CHECK=1 ./target/debug/vera overview 2>&1 >/dev/null | head -1   # this branch
    warning: index may be stale: 4 added, 33 modified. Search and grep only cover indexed files. Run `vera update .` or `vera watch .`.

No auto-indexing was added. `detect_staleness` walks and re-hashes every tracked
file, and indexing inside a tool call is a known problem on this surface.

## Cost, measured

A freshness scan against this repository's own index, release build, re-measured
at the head of this branch:

| call | elapsed |
|------|---------|
| 1 (cold) | 87.7 ms |
| 2 | 20.9 us |
| 3-5 | 5.0-7.3 us |

88 ms per tool call is not acceptable on a hot path, so the scan result is cached
for 30 seconds per repository (`STALENESS_TTL` in `crates/vera-mcp/src/staleness.rs:15`).
The trade is bounded in both directions: at most 30 seconds of a stale notice after
an update lands, or of silence after the tree drifts. The CLI does not cache,
because it scans once per process.

The cached path costs microseconds rather than nanoseconds because it resolves the
repository path before looking it up, so `.` and the absolute path to the same tree
share one entry. Against an 88 ms scan, and against the work a search tool call
does anyway, that is not a cost worth avoiding.

Up to eight repositories are cached at once (`CACHE_CAPACITY`). One entry was not
enough: `get_overview` takes a `path` argument, so a single server answers about
more than the tree it was started in, and checking a second repository evicted the
first and forced a rescan inside the TTL.

## Where it applies

`search_code`, `regex_search`, `structural_search`, `find_references` and
`get_overview` — the tools whose output is derived from the index. `get_overview`
is included because `vera overview` warns on the CLI side too.

A scan failure is logged at debug and treated as fresh, matching `warn_if_index_stale`.

## Tests

- `tools::tests::stale_index_result_carries_the_warning_after_the_results` — asserts
  the full notice text in `content[1]` and that `content[0]` is the unmodified
  results JSON.
- `tools::tests::fresh_index_result_is_a_single_results_block` — a matching index
  produces exactly one block.
- `staleness::tests::{stale_index_produces_a_notice_naming_the_drift, fresh_index_produces_no_notice, checking_one_repo_does_not_evict_another}`
- `protocol::tests::{tool_call_result_notice_is_a_trailing_block, tool_call_result_without_notice_has_one_block}`
- `indexing::freshness::tests::{stale_warning_names_every_kind_of_drift, fresh_index_has_no_stale_warning}`

Verified by reinjection. Reverting `results_with_staleness` to the old
`ToolCallResult::success(json)` and rerunning:

    ---- tools::tests::stale_index_result_carries_the_warning_after_the_results stdout ----
    assertion `left == right` failed: stale index must add a notice block, got [TextContent { type: "text", text: "[{\"file_path\":\"src/lib.rs\"}]" }]
      left: 1
     right: 2

    test result: FAILED. 43 passed; 1 failed

And, for the per-repository cache, restoring the single-slot cache:

    ---- staleness::tests::checking_one_repo_does_not_evict_another stdout ----
    assertion `left == right` failed: a second repository's check must not evict the first repository's cached scan
      left: None
     right: Some("warning: index may be stale: 1 modified. Search and grep only cover indexed files. Run `vera update .` or `vera watch .`.")

    test result: FAILED. 3 passed; 1 failed

## Verification

    cargo fmt --check                                            clean
    cargo test -p vera-core --lib                                795 passed, 0 failed
    cargo test -p vera-cli --bin vera                             97 passed, 0 failed
    cargo test -p vera-mcp                                        44 passed, 0 failed
    cargo build --workspace                                      clean
    cargo clippy -p vera-core --lib | grep -cE "^warning: [a-z]"   5   (all pre-existing)

`cargo clippy -p vera-mcp --lib --tests` and `cargo clippy -p vera-cli --bin vera`
add no warnings of their own; both report only the same five pre-existing
`vera-core` ones.

## Not in this PR

Issue #117 has three parts. This is the first. Still open:

- the watcher rebuilding models on every debounce tick;
- `vera serve` loading the index per request.

Both are separate changes and are not touched here.

### DIFF
--- a/crates/vera-cli/src/helpers.rs
+++ b/crates/vera-cli/src/helpers.rs
@@ -66,16 +66,13 @@ pub fn load_runtime_config() -> anyhow::Result<vera_core::config::VeraConfig> {
 
 pub fn warn_if_index_stale(repo_path: &Path, indexing_config: &vera_core::config::IndexingConfig) {
     match vera_core::indexing::detect_staleness(repo_path, indexing_config) {
-        Ok(freshness) if freshness.is_stale() => {
-            let stderr = std::io::stderr();
-            let mut err = stderr.lock();
-            let _ = writeln!(
-                err,
-                "warning: index may be stale: {}. Search and grep only cover indexed files. Run `vera update .` or `vera watch .`.",
-                freshness.summary()
-            );
+        Ok(freshness) => {
+            if let Some(warning) = freshness.stale_warning() {
+                let stderr = std::io::stderr();
+                let mut err = stderr.lock();
+                let _ = writeln!(err, "{warning}");
+            }
         }
-        Ok(_) => {}
         Err(err) => {
             tracing::debug!(error = %err, "failed to check index freshness");
         }
--- a/crates/vera-core/src/indexing/freshness.rs
+++ b/crates/vera-core/src/indexing/freshness.rs
@@ -48,6 +48,19 @@ impl IndexFreshness {
         }
         parts.join(", ")
     }
+
+    /// The stale-index warning shown to users, or `None` when the index covers
+    /// the tree. Single-sourced here because the CLI prints it on stderr and
+    /// the MCP server attaches it to tool results, and the two must agree.
+    pub fn stale_warning(&self) -> Option<String> {
+        if !self.is_stale() {
+            return None;
+        }
+        Some(format!(
+            "warning: index may be stale: {}. Search and grep only cover indexed files. Run `vera update .` or `vera watch .`.",
+            self.summary()
+        ))
+    }
 }
 
 pub(crate) fn record_index_snapshot(
@@ -216,6 +229,26 @@ mod tests {
             std::fs::create_dir_all(parent).unwrap();
         }
         std::fs::write(path, content).unwrap();
+    }
+
+    #[test]
+    fn stale_warning_names_every_kind_of_drift() {
+        let freshness = IndexFreshness {
+            files_added: 1,
+            files_modified: 2,
+            files_deleted: 3,
+        };
+        assert_eq!(
+            freshness.stale_warning().unwrap(),
+            "warning: index may be stale: 1 added, 2 modified, 3 deleted. \
+             Search and grep only cover indexed files. \
+             Run `vera update .` or `vera watch .`."
+        );
+    }
+
+    #[test]
+    fn fresh_index_has_no_stale_warning() {
+        assert_eq!(IndexFreshness::default().stale_warning(), None);
     }
 
     #[test]
--- a/crates/vera-mcp/src/lib.rs
+++ b/crates/vera-mcp/src/lib.rs
@@ -16,5 +16,7 @@ pub mod protocol;
 pub mod saved_config;
 pub mod server;
 pub mod staleness;
+#[cfg(test)]
+mod test_support;
 pub mod tools;
 pub mod watcher;
--- a/crates/vera-mcp/src/lib.rs
+++ b/crates/vera-mcp/src/lib.rs
@@ -15,5 +15,6 @@
 pub mod protocol;
 pub mod saved_config;
 pub mod server;
+pub mod staleness;
 pub mod tools;
 pub mod watcher;
--- a/crates/vera-mcp/src/protocol.rs
+++ b/crates/vera-mcp/src/protocol.rs
@@ -156,6 +156,22 @@ impl ToolCallResult {
         }
     }
 
+    /// Create a success result carrying an optional advisory notice.
+    ///
+    /// The notice is a separate trailing content block, so `content[0]` stays
+    /// byte-identical to what [`ToolCallResult::success`] would have produced
+    /// and clients that parse it as JSON are unaffected.
+    pub fn success_with_notice(text: impl Into<String>, notice: Option<String>) -> Self {
+        let mut result = Self::success(text);
+        if let Some(notice) = notice {
+            result.content.push(TextContent {
+                r#type: "text",
+                text: notice,
+            });
+        }
+        result
+    }
+
     /// Create an error result with text content.
     pub fn error(text: impl Into<String>) -> Self {
         Self {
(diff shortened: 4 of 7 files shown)


citron07r:
The CodeRabbit check on this PR reports `pass` with the reason **`Review rate limited`**, so the hosted review never actually read the diff. I ran the review locally instead, from the branch's worktree:

    coderabbit review --committed --base-commit e3d79b3b313c7a5553bde60cc22780c83667324f

It returned two findings. Both are addressed; what follows is what each claim turned out to be when checked by execution rather than by reading.

## Fixed: the staleness cache was a single global slot (major)

> The singleton cache evicts repository A when repository B is checked. A later request for A rescans it before its 30-second TTL expires.

True, and reachable on this surface: `get_overview` takes a `path` argument, so one server answers about more than the tree it was started in. Proven before fixing, with a test that checks repo A, then repo B, then makes A's tree genuinely fresh on disk so that a rescan would answer `None`:

    ---- staleness::tests::checking_one_repo_does_not_evict_another stdout ----
    assertion `left == right` failed: a second repository's check must not evict the first repository's cached scan
      left: None
     right: Some("warning: index may be stale: 1 modified. Search and grep only cover indexed files. Run `vera update .` or `vera watch .`.")

    test result: FAILED. 3 passed; 1 failed

The cache is now keyed by resolved path and bounded at eight repositories, and the entry survives. Resolving the path also means `.` and the absolute path to the same tree stop being two entries.

## Fixed, with a different rationale: the lock held across the scan

> The lock also remains held while `detect_staleness` re-hashes tracked files, so a scan for one repository blocks all other repositories.

The stated harm does not exist here. `run_server` (`crates/vera-mcp/src/server.rs:32`) is a blocking `read_line` loop over stdio, so tool calls are already serialized; the crate's one background thread (`watcher.rs:103`) runs `run_incremental_update` and never touches this cache. There is nothing for the lock to block. I released the lock across the scan anyway, because holding a mutex across a walk that re-hashes every tracked file is worth avoiding on its own, and a duplicated concurrent scan of the same repository is wasteful rather than wrong.

## Fixed: duplicated test fixture (trivial)

> `repo_indexed_as` repeats the temporary repository, `.vera/metadata.db`, and content-hash setup in `crates/vera-mcp/src/staleness.rs`.

Correct. Lifted into one `test_support` helper used by both modules.

## Found separately, not by the bot: the notice text was copied, not shared

The body of this PR claimed "wording matches `warn_if_index_stale` verbatim, so the CLI and MCP surfaces agree". That was a convention, not a mechanism: the format string existed twice, once per crate, and nothing would have caught an edit to one of them. It is now `IndexFreshness::stale_warning` in `vera-core`, called by both. The CLI's output is unchanged, checked against the released binary on the same repository at the same moment:

    $ VERA_NO_UPDATE_CHECK=1 vera overview 2>&1 >/dev/null | head -1        # vera 1.0.0
    warning: index may be stale: 4 added, 33 modified. Search and grep only cover indexed files. Run `vera update .` or `vera watch .`.
    $ VERA_NO_UPDATE_CHECK=1 ./target/debug/vera overview 2>&1 >/dev/null | head -1   # this branch
    warning: index may be stale: 4 added, 33 modified. Search and grep only cover indexed files. Run `vera update .` or `vera watch .`.

## Corrections to the description

Re-derived at the current head rather than carried over:

- The cost table was re-measured. Cold 87.7 ms, second call 20.9 us, then 5.0-7.3 us. The cached path is microseconds rather than the ~125 ns previously claimed, because it now resolves the repository path before the lookup. Against an 88 ms scan that is not a cost worth avoiding, and the reason is stated in the body.
- `helpers.rs:186` is now `helpers.rs:183`; `tools.rs:632` is labelled as the line at the base commit.
- Test names and counts updated: `staleness::tests::notice_reports_deletions` moved to `vera-core` as `indexing::freshness::tests::stale_warning_names_every_kind_of_drift`, alongside `fresh_index_has_no_stale_warning`.

## Verification, at the current head

    cargo fmt --check                                            clean
    cargo test -p vera-core --lib                                795 passed, 0 failed
    cargo test -p vera-cli --bin vera                             97 passed, 0 failed
    cargo test -p vera-mcp                                        44 passed, 0 failed
    cargo build --workspace                                      clean
    cargo clippy -p vera-core --lib | grep -cE "^warning: [a-z]"   5   (all pre-existing)

`cargo clippy -p vera-mcp --lib --tests` and `cargo clippy -p vera-cli --bin vera` add none of their own.

Issue #106, opened 2026-08-20
Rust inline mod swallows every symbol inside it: functions are absent from the call graph


citron07r:
Every symbol defined inside a Rust inline `mod { … }` is swallowed by the module symbol. Functions inside it do not exist in the call graph, and callers inside it are attributed to the module rather than the calling function.

This is wrong on the happy path, always, with no error anywhere. It affects most Rust files in most Rust repos, including Vera's own, because `#[cfg(test)] mod tests { … }` is the standard idiom.

## Reproduced against the released binary, on Vera's own index

Caller attribution. `tier0_line_chunks` is called from test functions inside `mod tests`:

````
$ vera references tier0_line_chunks
```crates/vera-core/src/parsing/chunker.rs:644-648 module:tests (part 1)
````

Every hit is `module:tests (part 1)`. The function that actually makes the call is not recorded.

Callee direction, for a function *defined* inside the inline mod:

````
$ vera references tier0_fallback_produces_chunks --callees
No callees found for 'tier0_fallback_produces_chunks'.
````

It has callees. It calls `tier0_line_chunks`. The symbol is simply absent from the graph.

The source is indexed, so this is extraction, not discovery:

````
$ vera grep "fn tier0_fallback_produces_chunks"
```crates/vera-core/src/parsing/chunker.rs:637-641
````

## Cause

`crates/vera-core/src/parsing/extractor/classify.rs:444` maps `mod_item` to `SymbolType::Module`, and `parsing/extractor/mod.rs` has explicit recursion cases so that container nodes do not swallow their children:

| container | site |
|---|---|
| Rust `impl_item` | `extractor/mod.rs:283` |
| Python `class_definition` | `extractor/mod.rs:290` |
| C# `namespace_declaration` | `extractor/mod.rs:326` |

There is **no case for `mod_item`**. The node is pushed as a single symbol spanning the whole block and extraction returns without descending.

Verified on `master` at `e3d79b3`.

## Consequences

- `vera references <fn>` inside an inline mod returns nothing, indistinguishable from a genuinely uncalled function.
- `vera references <fn> --callers` attributes calls to the module, so call-graph navigation stops at the module boundary.
- `vera dead-code` cannot see these functions as callers, so anything only called from inside an inline mod looks dead.
- `vera search --type function` cannot find them.
- The chunk spans the whole module, so one embedding represents an entire test module rather than individual functions, which also degrades retrieval quality for anything inside it.

## Fix shape

[...]

PR #131, opened 2026-08-20
fix(parsing): stop body-spanning containers swallowing their members

citron07r:
Follow-up to #106. #126 fixes this for Rust; this extends the same fix to the other fourteen languages that share it.

## The defect

`collect_symbols` stops at the first node `classify_node` recognises. When that node is a container whose span covers a body (a class, module, namespace, trait, object), every declaration inside it is inside that span and none of them becomes a symbol of its own. The result is one chunk for the whole container and no symbol for any method in it, so `references`, `dead-code` and symbol-scoped search cannot see them.

`crates/vera-core/src/parsing/extractor/mod.rs` already had one case for this shape, C# `namespace_declaration`: record the container, then keep walking. Fourteen more languages needed it.

## Verification, per language, by parsing

Each row was confirmed by parsing a snippet and reading the extracted symbols, not by reading the classification table. Before, only the container came out; after, the container plus its members.

| Language | Container kinds | Was swallowed |
|---|---|---|
| Ruby | `class`, `module` | nested class, `method`, `singleton_method` |
| Kotlin | `class_declaration`, `object_declaration` | `function_declaration` |
| Swift | `class_declaration` | `function_declaration` |
| Scala | `class_definition`, `trait_definition`, `object_definition` | `function_definition`, nested `class_definition` |
| C++ | `namespace_definition`, `class_specifier`, `struct_specifier` | `class_specifier`, `function_definition` |
| HLSL | `class_specifier`, `struct_specifier` | `function_definition` |
| Groovy | `class_definition`, `class_declaration`, `interface_definition`, `interface_declaration` | `method_declaration` |
| PowerShell | `class_statement` | `class_method_definition` |
| MATLAB | `class_definition` | `methods` > `function_definition` |
| D | `class_declaration`, `struct_declaration`, `interface_declaration`, `template_declaration` | `aggregate_body` > `function_declaration` |
| OCaml | `module_definition` | `value_definition`, `type_definition` |
| F# | `module_defn` | `function_or_value_defn`, `type_definition` |
| Julia | `module_definition` | `function_definition`, `struct_definition`, nested `module_definition` |
| Fortran | `module`, `program` | `internal_procedures` > `function`/`subroutine` |

Four of these are wider than a first reading of the table suggests, and the parse is why:

- **C++ `struct_specifier`** swallows inline member definitions exactly as `class_specifier` does, so it is included. C's `struct_specifier` is not: C structs hold no callables, and a sibling function after one is already extracted (`[Struct Some("Holder") @0, Function Some("helper") @4]`).
- **HLSL** is a C++-family grammar and its table classifies the same two kinds. `class Widget { int render() { return 1; } };` extracted only `Class Some("Widget")`. GLSL is not included: GLSL has no member functions, so its `struct_specifier` holds nothing to swallow. Covering C++ and not HLSL would have left two spellings of the same construct behaving differently.
- **D `template_declaration`** holds `function_declaration` children through the same `aggregate_body`, so it is a container too. D's `module_declaration` is *not*: it spans one line and is already a leaf, which a test pins.
- **Swift `class_declaration`** is also how the grammar spells `struct`, `enum` and `extension`, so one kind covers four constructs. The refinement step above the new case still runs first, so a struct is recorded as `Struct` and an enum as `Enum`, not flattened to `Class`. An `extension` carries no name the extractor can read and stays `Class None`, which is what it was before.

Groovy lists both the `_definition` and `_declaration` spellings because `GROOVY_KINDS` directly above does. Only the `_declaration` pair is reachable in the vendored grammar; keeping the mirror avoids the two tables diverging on a grammar bump.

Not changed, confirmed by the same method:

- **Header nodes that never span a body.** Perl `package_statement`, Erlang `module_attribute`, Elm `module_declaration`, Clojure `ns`, Common Lisp `defpackage` all end at the declaration; the functions after them are already extracted as siblings. Parsed: Perl gives `[Module None @0, Function Some("bar") @1]`, Clojure `[Module Some("foo") @0, Function Some("bar") @1]`, and so on.
- **C# `file_scoped_namespace_declaration`.** It is listed, because the case this replaces listed it, but it is a header: it ends at its semicolon (`1..15` for `namespace App;`) and recursing into it finds nothing. A test pins that its members are siblings rather than children, so nobody later "fixes" a nesting assertion that never held.
- **Whole-unit chunks.** CSS `rule_set`, GraphQL type definitions hold nothing the extractor classifies. HTML/XML `element`, INI `section`, Nix `let_expression` and HCL `block` do span classified children, but those children are markup and data rather than callables.
- **Two constructs that do swallow and are still left alone**, recorded so the boundary of this change is explicit rather than an oversight. Protobuf `message` swallows a nested `message` and `enum` (`message Outer { ... message Inner { ... } }` extracts `[Struct Some("Outer") @1]` only). CMake `if_condition`/`foreach_loop`/`while_loop` swallow a `function_def` defined inside them (`[Block None @0]` only). Neither is a class-like container of callables, and both change chunking for constructs that appear in almost every file of their language, so they want their own change and their own measurement.
- **OCaml `class_definition` and `module_type_definition`** are recursed into by nothing, because they span only `method_definition` and `value_specification` and the OCaml table classifies neither. Parsed: `class widget = object method render = 1 end` gives `[Class None @0]` with or without an entry, so adding one would be a no-op.

## Two supporting changes the recursion needs

**Anonymous nodes are now skipped** (`extractor/mod.rs:138`). Ruby spells the class keyword `class` and Fortran spells it `module` — the same node kinds those tables classify. Walking into the container therefore recorded its own keyword token as a second `Class`/`Module` symbol. Anonymous nodes are punctuation and keywords and are never symbols; this was latent before only because classified containers returned early and their keyword children were never visited.

The guard is a change to every language, so its blast radius was measured rather than argued. Sweeping all 61 grammar-backed languages for anonymous node kinds that `classify_node` maps to a `SymbolType` returns sixteen (language, kind) pairs. Thirteen also exist as a *named* kind, so the guard only ever skips their keyword token. The other three are Common Lisp `defmacro`, `defgeneric` and `defmethod`, which the grammar wraps in a named `defun` node handled higher up, so they are unreachable either way — `(defmacro m (x) x)` extracts `Function Some("m")` spanning the whole form with the guard in place and without it. No other language has an anonymous kind that classifies at all.

**Two table corrections** (`extractor/classify.rs`):
- PowerShell `class_method_definition` is classified as `Method`. Without it, recursing into a PowerShell class finds nothing to record and the fix is a no-op for that language.
- Fortran `module_statement` and `program_statement` are no longer classified. They are the name-bearing header lines of the `module`/`program` node that is now recorded itself, so classifying them yields a duplicate symbol spanning only the header. They were unreachable before this change, because `module`/`program` returned early.

## Tests

`crates/vera-core/src/parsing/extractor/container_tests.rs`, one fixture per language, sixteen tests. Each asserts presence before absence and pins identity rather than counts: the inner symbols are required by name and `SymbolType` (by `SymbolType` and start row where the grammar gives the extractor no name to read), the container is required too, and each inner symbol is asserted to sit strictly inside the container's span. Failures print the whole extracted symbol list.

Every entry in `container_body_kinds` was checked for a fixture that discriminates, by deleting that one entry and rerunning. Each of C# `namespace_declaration`, C++ `struct_specifier`, HLSL's pair, D `class_declaration`, `struct_declaration`, `interface_declaration` and `template_declaration`, and Fortran `program` fails exactly one test and leaves the rest green, for example:

    ---- parsing::extractor::container_tests::hlsl_class_and_struct_keep_their_members ----
    expected Function "render", got [Class Some("Widget") @1, Struct Some("Holder") @5]

    ---- parsing::extractor::container_tests::cpp_namespace_and_class_keep_their_members ----
    expected Function "fetch", got [Module None @1, Class Some("Widget") @3, Function Some("render") @5, Function Some("helper") @8, Struct Some("Holder") @12]
    test result: FAILED. 15 passed; 1 failed

Verified by reinjection in the other directions too:

- **Case reverted, all languages:** replacing the whole table with `_ => &[]` fails all sixteen with the swallowing symptom (`test result: FAILED. 0 passed; 16 failed`), for example `expected Function "render", got [Class Some("Widget") @1, Class Some("Registry") @5]` (Kotlin), `expected Function at row 3, got [Module None @1]` (Fortran) and `expected Class "Widget", got [Module Some("App") @1]` (C#).
- **Anonymous-node guard reverted:** Ruby and Fortran fail on the duplicate, and only they do — the other 13 stay green and the full `vera-core` suite reports exactly those two failures:

      ---- ruby_module_and_class_keep_their_members ----
      expected one Class, also got None @2: [Module Some("Outer") @1, Module None @1, Class Some("Widget") @2, Class None @2, Function Some("render") @3, Function Some("helper") @8]

      ---- fortran_module_keeps_its_procedures ----
      expected one Module, also got None @1: [Module None @1, Module None @1, Function None @3]

      test result: FAILED. 805 passed; 2 failed

- **Extraction broken to empty:** all fail with `got []`, so none of them can pass on an extractor that returns nothing.

No existing test asserted the swallowing behaviour, so none had to be updated.

## Verification

    cargo fmt --check                                            clean
    cargo test -p vera-core --lib                                809 passed, 0 failed
    cargo test -p vera-cli --bin vera                             97 passed, 0 failed
    cargo build --workspace                                      clean
    cargo clippy -p vera-core --lib | grep -cE "^warning: [a-z]"   5   (all pre-existing)

## What this does not reach

`collect_symbols` stops at depth 6, and recursing into a container now spends that budget. One level of nesting is fine; two is not. Parsed at the head of this branch:

    namespace a { class C { int m() {...} }; }              [Module None @0, Class Some("C") @1, Function Some("m") @3]
    namespace a { namespace b { class C { int m() {...} }; } }  [Module None @0, Module None @1, Class Some("C") @2]
    module A; module B; class C; def m ... (Ruby)           [Module Some("A") @0, Module Some("B") @1, Class Some("C") @2]

The method is missing in the last two. That is not a regression — before this change it was missing at every depth, together with the class — but it is the boundary of the fix, and raising the limit is a separate change with its own chunking consequences.

## Not fixed here

Julia function names are wrong, and this change makes more of them visible rather than causing them. `function f(x)\n  x\nend` extracts `Function Some("x")` at the top level and `Function Some("x")` inside a module; `function f(x)\n  x + 1\nend` extracts `Function None`. The name comes from `extract_name`, which this PR does not touch, and it is equally wrong outside any container. It wants its own issue.

## Re-indexing

This changes what `extract_symbols` returns, so every file in these languages chunks differently. **A full re-index is required, and nothing detects that it is needed**: `model_identity` covers only the embedding model, and freshness is a per-file content hash, so unchanged files are skipped. Verified in #126 by running a new binary against an index built by an old one and getting "Index is up to date". Users on these languages need `vera index .` explicitly.

## Relationship to #126

Cut from master, so #126's Rust fix is not present here. The two touch the same region of `collect_symbols` and will conflict textually: #126 adds a `mod_item` case immediately after the Rust `impl_item`/`trait_item` case, this PR replaces the C# namespace case a few lines below with the table-driven one. The conflict is adjacency only, not semantics; whichever merges second rebases cleanly by keeping both blocks.

Once both are in, Rust's `mod_item` is the same shape as the new table entries and would fold into `container_body_kinds` as `Language::Rust => &["mod_item"]`, deleting #126's dedicated block. Not done here, so as not to pre-emptively refactor code still in review. Rust `impl_item`/`trait_item` stay on their own path either way: they push `Method` for their `function_item` children, which the generic recursion would record as `Function`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved code symbol extraction across many programming languages, including Ruby, Kotlin, Swift, C#, Scala, C++, PowerShell, Fortran, and others.
  - Nested classes, modules, structs, interfaces, traits, templates, functions, methods, and type aliases are now identified more reliably.
  - Improved handling of unnamed containers, file-scoped namespaces, header-only declarations, and nested symbol boundaries.
  - Added support for recognizing Fortran modules and programs, plus PowerShell class methods.
  - Prevented duplicate symbols during extraction.


citron07r:
The CodeRabbit check on this PR reports `pass` with the reason **`Review rate limited`**, so the hosted review never read the diff. I ran the review locally instead, from the branch's worktree:

    coderabbit review --committed --base-commit e3d79b3b313c7a5553bde60cc22780c83667324f

It returned one finding, on `crates/vera-core/src/parsing/extractor/mod.rs:13-17`.

## Fixed: test module declarations belong at the file bottom (major)

> Move both `#[cfg(test)]` module declarations after the production implementation. As per path instructions, `crates/`: "tests in `#[cfg(test)]` modules at file bottom."

The declaration this PR added sat next to a pre-existing one at the top of the file, so the immediate neighbour argued for leaving it. The wider count does not: of the four other `vera-core` modules that declare a test module as a separate file, three put the declaration at the bottom (`parsing/mod.rs` 233/236, `embedding/mod.rs` 56/56, `local_models/mod.rs` 498/597) and one does not (`retrieval/ranking/mod.rs` 21/132). Both declarations moved, rather than splitting the pair.

That was the whole of the bot's output. Everything below is from checking the change's own claims, which is where the substance was.

## Found separately, by checking the change's own claims

Three gaps that the description asserted and no assertion covered, each proven by execution before fixing.

### HLSL had the same bug and was in neither list

HLSL's table classifies `class_specifier` and `struct_specifier`, the same two kinds this PR added for C++, and its grammar is the same C++ family. Parsed at the base commit:

    class Widget { int render() { return 1; } };
    struct Holder { int fetch() { return 2; } };
    => [Class Some("Widget") @0, Struct Some("Holder") @3]

Both methods swallowed. Covering C++ and not HLSL in one commit would have left two spellings of the same construct behaving differently, so HLSL is now in the table with its own fixture. GLSL stays out: it has no member functions, so its `struct_specifier` holds nothing to swallow.

### Five table entries had no fixture that could fail

`container_body_kinds` could have lost C# `namespace_declaration` and `file_scoped_namespace_declaration`, C++ `struct_specifier`, D `interface_declaration` and `template_declaration`, or Fortran `program` with all thirteen tests still green. Each now has one, and each was checked by deleting that single entry and confirming exactly one test fails, for example:

    ---- container_tests::cpp_namespace_and_class_keep_their_members ----
    expected Function "fetch", got [Module None @1, Class Some("Widget") @3, Function Some("render") @5, Function Some("helper") @8, Struct Some("Holder") @12]
    test result: FAILED. 15 passed; 1 failed

Writing the C# fixture turned up a real correction: `file_scoped_namespace_declaration` is not a container at all. `namespace App;` spans bytes `1..15` and stops, so its members are siblings, not children, and recursing into it finds nothing. It stays in the list because the case this PR replaces listed it, but the doc comment claiming every listed kind spans a body was wrong and now says so, and the test pins siblings rather than a nesting that never held.

The Swift fixture also only covered `class` and `struct` while the description claimed one kind covers four constructs; it now covers `enum` and `extension` too.

### The anonymous-node guard was argued, not measured

`if !node.is_named() { return; }` changes every language, and the description justified it from Ruby and Fortran alone. Sweeping all 61 grammar-backed languages for anonymous node kinds that `classify_node` maps to a `SymbolType` returns sixteen (language, kind) pairs: Ruby `class`/`module`, Fortran `module`/`subroutine`/`function`/`program`, Protobuf `enum`/`message`/`service`/`rpc`, Common Lisp `defun`/`defmacro`/`defgeneric`/`defmethod`, Haskell `newtype`, OCaml `external`.

Thirteen of the sixteen also exist as a *named* kind, so the guard only ever skips their keyword token. The three that do not — Common Lisp `defmacro`, `defgeneric`, `defmethod` — are unreachable either way, because the grammar wraps all three in a named `defun` node that `collect_symbols` handles before any recursion reaches the keyword. Checked by execution: `(defmacro m (x) x)` extracts `Function Some("m")` spanning the whole form both with the guard and with it disabled. Running the whole `vera-core` suite with the guard disabled fails exactly two tests, the Ruby and Fortran duplicates, and nothing else.

## Also recorded, not fixed

- **Protobuf `message` and CMake's conditional blocks swallow too.** `message Outer { ... message Inner { ... } }` extracts `[Struct Some("Outer") @1]`; a `function_def` inside a CMake `if_condition` extracts `[Block None @0]`. Neither is a class-like container of callables and both would change chunking for constructs that appear in nearly every file of their language, so they want their own change. Written into the doc comment so the boundary of this one is explicit.
- **The depth-6 ceiling limits the fix to one level of nesting.** `namespace a { namespace b { class C { int m() {...} }; } }` still loses `m`; so does a Ruby module inside a module inside a class. Not a regression — before this change `m` was missing at every depth, along with its class — but it is where the fix stops, and raising the limit is a separate change.
- **Julia function names are wrong**, and this change makes more of them visible rather than causing them. `function f(x)\n  x\nend` extracts `Function Some("x")`, at the top level and inside a module alike; `function f(x)\n  x + 1\nend` extracts `Function None`. `extract_name` is untouched here and is equally wrong outside any container. It wants its own issue.

## Verification, at the current head

    cargo fmt --check                                            clean
    cargo test -p vera-core --lib                                809 passed, 0 failed
    cargo test -p vera-cli --bin vera                             97 passed, 0 failed
    cargo build --workspace                                      clean
    cargo clippy -p vera-core --lib | grep -cE "^warning: [a-z]"   5   (all pre-existing)

Reinjection at the current head, all three directions: replacing the whole table with `_ => &[]` fails all sixteen container tests (`0 passed; 16 failed`) with the swallowing symptom; disabling the anonymous-node guard fails exactly Ruby and Fortran on the duplicate (`805 passed; 2 failed`); making `extract_symbols` return an empty vec fails all sixteen with `got []`.

PR #132, opened 2026-08-20
perf(retrieval): build the reranker lazily, only for queries that rerank

### LINKED ISSUE #100: Reranker is loaded on every search, including searches that skip reranking (1.5 GB measured)
`SearchContext::new` builds the reranker unconditionally, before the query is known. For queries that skip reranking, that is roughly a gigabyte of ONNX session built and never used.

## Measured

Apple M4, released `vera 1.0.0`, same 375-file / 4,790-chunk index, peak RSS via `/usr/bin/time -l`:

| command | peak RSS | reranker inferences |
|---|---|---|
| `vera grep "WRITER_HEAP_SIZE"` | **16.4 MB** | n/a, no models loaded |
| `vera search "Bm25Index"` | **1.541 GB** | **zero** |
| `vera search "how are embeddings batched during indexing"` | 2.216 GB | runs |

So an identifier query pays ~1.52 GB for two ONNX sessions, and never invokes one of them. Model weights are 672 MB on disk (405 MB embedding + 267 MB reranker), so the loaded footprint is about 2.3x the file size.

## Cause

`crates/vera-core/src/retrieval/search_service.rs:89-94`, inside `SearchContext::new`:

````rust
let reranker = if provider.is_some() {
    crate::retrieval::create_dynamic_reranker(config, backend).await ...
````

`create_dynamic_reranker` builds and commits the full ONNX session eagerly (`retrieval/local_reranker.rs:22-74`, `:253-265`).

The decision to *use* it happens later, at `search_service.rs:217`, via `reranking_enabled` → `should_skip_reranking` (`:379-388`), which skips for short identifier queries, path-weighted queries, `exact_paths` and `symbol_type` filters. By then the session is already built.

The repo's own test suite pins the wasted case: `search_service.rs:695` asserts `!reranking_enabled(true, false, "authenticate", &filters)`. That is exactly the query shape measured above.

Verified on `master` at `e3d79b3`.

## Why it matters beyond RSS

Session construction is also latency on every search that skips reranking, and it is paid before any work starts. On memory-constrained machines it is the difference between `vera search` fitting and not.

## Suggested fix

Hold the reranker behind a `tokio::sync::OnceCell` and build it inside the `if reranking_enabled` branch. The codebase already has this shape: `vera-serve` caches providers lazily with idle eviction (`vera-serve/src/lib.rs:47-50`).

I have not measured the after case, so I am not claiming a number for the saving. Before sending a PR I would want per-query-class before/after RSS on a real index, and `.coderabbit.yaml` asks for Semble evidence on `retrieval/**` anyway.

Happy to do that work if you agree the eager build is worth changing.

### PR DESCRIPTION
**Pull request by citron07r**
`SearchContext::new` built the cross-encoder ONNX session before the query existed. The decision to *use* it is a per-query heuristic evaluated ~130 lines later, and it skips short identifier queries, path-weighted queries, and `exact_paths` / `symbol_type` filtered lookups. Those searches paid for a model they never called.

Fixes #100

## Change

`crates/vera-core/src/retrieval/search_service.rs`

- `SearchContext.reranker` is now `tokio::sync::OnceCell<Option<DynamicReranker>>` plus the `backend` needed to build it (`:66-74`), instead of an `Option<DynamicReranker>` filled in `new`.
- `SearchContext::reranker_for_query` (`:129-140`) applies the skip heuristic first and only then resolves the session; `SearchContext::reranker` (`:163-178`) does the `get_or_init`.
- `search` calls it in place of the old eager field read (`:282-285`).
- `reranking_enabled(reranker_present, has_intent, query, filters)` split into `reranking_wanted(has_intent, query, filters)` (`:446-448`). The availability half is now the `Option` returned by the lazy resolve, so the predicate that gates the build no longer needs a built reranker to answer.

This follows the lazy-provider shape `vera-serve` already uses (`crates/vera-serve/src/lib.rs:47-50`) rather than introducing a new pattern.

### Caching and failure handling

Laziness is per context, not per query. MCP keeps one `SearchContext` for the life of the session (`crates/vera-mcp/src/tools.rs:22-27`), so a long session still loads the session exactly once.

A failed build is cached for the life of the context and warned once, with the same message as before. The alternative, retrying per query, turns one absent model asset or unusable execution provider into a multi-second stall on every subsequent search, and those causes are static within a process. The cache is per context, so a new process or a rebuilt context retries; nothing is marked absent process-wide.

The existing warn-and-degrade path is unchanged: a reranker that fails to build yields `None`, and the search runs unreranked hybrid exactly as it did when the failure happened up front.

## Measured

Apple M4, macOS 15.6. Scratch index of 73 files / 2,675 chunks, built with released `vera 1.0.0`. Peak RSS via `/usr/bin/time -l`, best of three runs, same index and same query for both binaries.

| query | released 1.0.0 | this branch | delta |
|---|---|---|---|
| `Bm25Index` (skips reranking) | 1.534 GB | **0.780 GB** | **-754 MB, -49%** |
| `how are embeddings batched during indexing` (reranks) | 2.468 GB | 2.485 GB | +17 MB, +0.7% (run-to-run noise) |

Wall clock on the identifier query, three runs each:

| binary | real |
|---|---|
| released 1.0.0 | 0.76 / 0.73 / 0.76 s |
| this branch | 0.29 / 0.29 / 0.28 s |

So the identifier-query saving is about 750 MB of peak RSS and 0.45 s of latency, and the reranking case is unchanged within noise. The issue deliberately claimed no figure; this is it.

## Evidence for the retrieval path instruction

`.coderabbit.yaml` asks for Semble benchmark evidence on `crates/vera-core/src/retrieval/**`. This change moves *when* the reranker is constructed. It does not touch fusion, ranking, candidate pools, or the reranker call site, and it does not change which reranker instance a reranking query gets. Relevance is therefore unchanged by construction rather than by measurement.

Offered as the direct check instead: identical result sets before and after, on the same index, across all three query classes.

```
$ diff <(vera search "$q" --limit 15) <(./target/release/vera search "$q" --limit 15)

IDENTICAL: Bm25Index                                      # skips reranking
IDENTICAL: how are embeddings batched during indexing     # reranks
IDENTICAL: vector search knn candidate pool               # reranks
```

Byte-identical output, so ordering and scores match too.

## Tests

`reranker_is_built_only_for_queries_that_rerank` (`search_service.rs:799-851`) asserts construction, not results, since the results are identical either way. A build counter on the context records how many times the `get_or_init` body ran:

- nothing is built by `SearchContext::new`
- `"Bm25Index"` resolves to no reranker and leaves the counter at 0
- a natural-language query builds it, counter 1
- a second natural-language query reuses it, counter still 1

The counter field is `#[cfg(test)]`, so release builds carry no extra state.

`reranking_runs_for_intent_searches_with_short_queries` keeps its issue #20 guard, retargeted at `reranking_wanted`.

`bm25_only` documents that its `InferenceBackend::Api` value is unused because `search` returns on the BM25-only path before any reranker is resolved. Nothing enforced that, so `bm25_only_search_fills_path_scoped_results_from_deeper_pool` (`:683-760`) now ends on the build counter. Its query is one `reranking_wanted` accepts, so the assertion discriminates.

### Reinjection

Both were verified by reinjection at the head commit.

`SearchContext::new` made to force the cell before returning, which is the original eager build:

```
thread 'retrieval::search_service::tests::reranker_is_built_only_for_queries_that_rerank' panicked at
crates/vera-core/src/retrieval/search_service.rs:812:9:
assertion `left == right` failed
  left: 1
 right: 0

test result: FAILED. 0 passed; 1 failed
```

The reranker resolve moved above the provider check in `search`:

```
thread 'retrieval::search_service::tests::bm25_only_search_fills_path_scoped_results_from_deeper_pool' panicked at
crates/vera-core/src/retrieval/search_service.rs:759:9:
assertion `left == right` failed: a context with no embedding provider must never construct a reranker
  left: 1
 right: 0
```

(Both line numbers are inside the reinjected file, which is a few lines longer than the committed one.)

## Verification

Run at the head commit.

| check | result |
|---|---|
| `cargo build --workspace` | clean |
| `cargo fmt --check` | clean |
| `cargo test -p vera-core --lib` | 794 passed, 0 failed |
| `cargo test -p vera-cli --bin vera` | 97 passed, 0 failed |
| `cargo test -p vera-mcp` | 37 passed, 0 failed |
| `cargo clippy -p vera-core --lib` warnings | 5, the pre-existing set, none added |

---

---
## Summary by cubic
Builds the cross-encoder reranker lazily and only for queries that actually rerank, reducing peak RSS and latency for identifier/path-filtered searches with no result changes. Previously `SearchContext::new` eagerly constructed the ONNX session; now it resolves on the first reranking query and is skipped otherwise. Fixes #100.

- Replace eager `SearchContext.reranker: Option<DynamicReranker>` with `backend` + `tokio::sync::OnceCell<Option<DynamicReranker>>`; nothing builds in `new`.
- Add `reranker_for_query` to apply the skip heuristic before any build; `search` resolves lazily inside the rerank branch.
- Split `reranking_enabled(...)` into `reranking_wanted(...)`; availability comes from the lazy resolve, not the predicate.
- Cache success and failure per `SearchContext`; warn once; fallback to hybrid search unchanged.
- Tests: assert zero builds in `new`, no build for skipped queries, single build then reuse; pin the BM25-only invariant (no provider → no reranker). Tests adopt the shared `EnvVarGuard` from `master`.
- Measured: identifier query peak RSS 1.534 GB → 0.780 GB (−754 MB) and wall time ~0.73 s → ~0.29 s; reranking queries unchanged within noise.

<sup>Written for commit 71c030c1a13820e22c773522531b3ee7c5a625dd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/132?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Performance Improvements**
  * Reranking is initialized only for searches that need it, reducing unnecessary processing.
  * Reranking setup is reused across eligible searches for faster subsequent results.

* **Search Reliability**
  * Searches automatically fall back to standard hybrid search when reranking is unavailable.
  * Explicit requests for reranking continue to be honored.

### CODE REVIEW

**Comment by citron07r** [comment]
## Local review (the hosted one was rate limited)

The CodeRabbit check on this PR reported `pass` with the reason **`Review rate limited`**, so nothing was actually reviewed. I ran the review locally from the branch's worktree instead:

```
coderabbit review --committed --base-commit e3d79b3
```

**1 finding, trivial.** No major or minor findings. Pushed b89cc0b in response, plus one thing the local run did not raise.

### Declined: extract the shared env save/set helper (trivial, Maintainability)

> `set_test_reranker_env` duplicates the structure of `set_test_embedding_env` at Lines 513-524. Only the key list and values differ.

The duplication is real and it is the module's existing convention. `set_test_embedding_env` is pre-existing, this PR does not otherwise touch it, and both helpers already share the one piece that matters, `restore_test_env`. Deduplicating means rewriting the pre-existing helper too, or leaving three shapes where there were two — either way it turns a perf diff into a test-helper refactor of code the change does not touch. Eight lines of mirrored setup is under the bar for that.

### Added anyway: an assertion for an invariant nothing enforced

Not raised by the local run; found re-reading the diff against the repo's own checklist. `bm25_only` carries a comment claiming its `InferenceBackend::Api` value is never used, because `search` returns on the BM25-only path before any reranker is resolved. That was true and nothing held it true — a later reorder would silently start building a cross-encoder for a context with no embedding provider.

`bm25_only_search_fills_path_scoped_results_from_deeper_pool` already drives `"how does dependency injection work"`, which `reranking_wanted` accepts, so one assertion on the build counter covers it. Verified by moving the resolve above the provider check:

```
thread 'retrieval::search_service::tests::bm25_only_search_fills_path_scoped_results_from_deeper_pool' panicked at
crates/vera-core/src/retrieval/search_service.rs:759:9:
assertion `left == right` failed: a context with no embedding provider must never construct a reranker
  left: 1
 right: 0
```

Also dropped "build reranker" from the module header, which described the eager construction this branch removes.

### Body corrections

Every line reference in the description was re-derived at the head commit; several were stale (the test was cited as `:783-848`, it is `:799-851`). The reinjection output is now quoted from a run at the head commit rather than an earlier revision, and the verification table gained `cargo build --workspace` and `cargo test -p vera-mcp`.

### Re-verified at b89cc0b

| check | result |
|---|---|
| `cargo build --workspace` | clean |
| `cargo fmt --check` | clean |
| `cargo test -p vera-core --lib` | 794 passed, 0 failed |
| `cargo test -p vera-cli --bin vera` | 97 passed, 0 failed |
| `cargo test -p vera-mcp` | 37 passed, 0 failed |
| `cargo clippy -p vera-core --lib` | 5 warnings, the pre-existing set, none added |

### On the `retrieval/**` benchmark instruction

The local run did not ask for Semble evidence, but the standing answer is in the description: this change moves *when* the reranker is constructed and touches no fusion, ranking, candidate-pool or reranker call-site code, so ranking cannot move. The evidence offered is byte-identical `vera search` output between the released 1.0.0 binary and this branch, on the same index, across a skipped query, a reranked query and a filtered query.

PR #133, opened 2026-08-20
perf(mcp): build the watcher's runtime and provider once, not per debounce cycle

citron07r:
Part of #117

Issue #117 lists three divergences between MCP/`serve` and the CLI. This PR fixes
one of them: the file watcher rebuilding the embedding model on every debounce
cycle, and watching directories that indexing never walks.

The missing staleness warning is in flight on a separate branch. The `vera serve`
per-request model load is still open and untouched here.

## What was wrong

`run_update_blocking` built a **new tokio runtime and a new ONNX session on every
debounce cycle**, then dropped both. `create_dynamic_provider` has no cache, so
there was nothing to reuse.

The only event filter was `.vera`. The watch itself is recursive over the whole
tree, so every directory the indexer skips still produced events:

1. an agent calls `search_code` once, which starts the watcher
2. the user runs `cargo build`
3. `target/` churns; each debounce window sees changes that are not under `.vera`
4. each window spawns a thread, builds a runtime, builds an ONNX session, walks
   discovery, finds nothing indexable, tears it down

That repeats for the whole build, and for every build after. It is visible only
at `debug` level, so the cost is silent.

On Linux the recursive watch also registers inotify watches for `target/` and
`node_modules/`. If `fs.inotify.max_user_watches` is exhausted, `start_watching`
fails and the MCP caller only `warn!`s, so the index silently stops updating for
the rest of the session.

## The fix

**Build the engine once.** The runtime, provider, resolved config and model name
move into an `EmbeddingUpdateEngine` (`crates/vera-mcp/src/watcher.rs:55`) behind
a `SharedEngine` cell (`:37`) owned by the watcher (`:170`). Construction is
deferred to the first cycle that needs it, so starting a watcher never loads a
model, and a construction failure stays a per-cycle error the next cycle can
retry rather than something that kills the watcher. Saved config is now read once
at watcher start instead of once per cycle, so a config edit mid-session is
picked up on the next watcher, not the next cycle.

Concurrency is unchanged: cycles still run on spawned threads and are still
serialised by the existing `updating` compare-exchange, so the shared provider is
used by one cycle at a time.

**Filter on discovery's own exclusions.** `ExclusionMatcher`
(`crates/vera-core/src/discovery/mod.rs:204`) is built from `build_overrides`,
the same function `discover_files` installs on its walker. The watcher builds one
at start and asks it per event (`crates/vera-mcp/src/watcher.rs:190`). No second
list of directory names exists to drift from the first: `target`, `node_modules`,
`dist`, `build`, `__pycache__`, `.venv`, `.git` and any `--exclude` patterns all
come from `IndexingConfig`.

Scope boundary: the matcher covers `default_excludes` and `--exclude` only. It
does not consult `.gitignore` or `.veraignore`, which `discover_files` also
applies through its walker. A gitignored directory therefore still triggers an
update cycle that finds nothing to do. That is the safe direction (a spurious
cycle, never a missed file) and it keeps the per-event cost to a glob match.

Paths outside the root are reported as *not* excluded. The matcher cannot
classify them against root-relative patterns, and treating them as excluded would
silently drop work.

**The index directory is filtered on its own.** `.vera` is *not* routed through
`ExclusionMatcher` (`crates/vera-mcp/src/watcher.rs:165`, `:192`). An update cycle
writes into it, so a watcher that reacts to those writes re-triggers itself
forever, and `ExclusionMatcher` follows `IndexingConfig`: `no_default_excludes`,
or a stored `default_excludes` that omits `.vera`, switches the guard off. Both
reach the watcher, since `load_saved_runtime_config` deserialises the whole
`IndexingConfig` out of `~/.vera/config.json`. The pre-branch `starts_with` check
on `index_dir` is kept, unconditional, and the exclusion matcher handles only the
directories indexing skips.

**`is_excluded` probes the path instead of assuming a directory.** The default
patterns are directory-only (`!target/`), and the first revision of this branch
passed `is_dir = true` for every path. A regular file named `build`, `target` or
`dist` — ordinary extensionless script names, which `discover_files` indexes
because it applies no extension allowlist — therefore matched, and the watcher
ignored edits to a file that is in the index. `crates/vera-core/src/discovery/mod.rs:232`
now calls `path.is_dir()`. A path that no longer exists is treated as a file so
deleting such a file still triggers an update; parents are matched as directories
either way, which keeps deletions under `target/` excluded. The one behaviour
this gives up is that removing the `target` directory itself now costs one
update cycle.

**OS watch limit.** `watch_failure_message`
(`crates/vera-mcp/src/watcher.rs:245`) turns `notify::ErrorKind::MaxFilesWatch`
into a message that names the limit, says auto-update is off, and gives the
remedy, instead of surfacing as a bare io error. It stays a `warn!` in the MCP
path on purpose: failing `search_code` because a watcher could not start would be
worse than serving the search. The retry is already there, since `tools.rs` only
stores a handle on success and so re-attempts on the next `search_code` call.

I did not restrict *which* directories get watched. Registering non-recursive
watches per non-excluded subtree needs the debouncer to add watches for
directories created later, which requires a self-referential handle inside the
event closure; without it, a newly created top-level directory would silently go
unwatched. Falling back to a poll watcher on `MaxFilesWatch` was also rejected:
periodically stat-walking a repository that has `target/` in it is exactly the
cost this PR removes.

## Why not share the provider with the MCP `SearchContext` cache

`tools.rs` already caches an `Arc<SearchContext>` in the same process, so reusing
it was the obvious first idea. It is the wrong provider for this call site:

- `SearchContext` exposes a `CachedEmbeddingProvider<DynamicProvider>`, a
  512-entry LRU over query text. Indexing pushes every chunk of every changed
  file through it, which would evict the query cache it exists to serve and add
  hashing and lock traffic per chunk for a hit rate near zero on document text.
- `SearchContext` also builds a reranker, which the update path never calls.
  Depending on it would tie index freshness to reranker construction.
- `watcher.rs` also backs `vera watch` in the CLI, which never enters `tools.rs`.
  A cache reachable only from the MCP tool dispatch would leave the CLI watcher
  rebuilding per cycle, which is the bug.

Lifetime: the engine is owned by the `SharedEngine` the watcher's event closure
holds, so it lives exactly as long as the watcher, and each cycle takes an `Arc`
clone for the duration of its update. The MCP process therefore holds at most two
embedding providers, one for search and one for indexing, each built once,
against one per debounce cycle before.

## Tests

Three in `crates/vera-mcp/src/watcher.rs`, driving real filesystem events through
a real debouncer with a counting stub engine and a 300ms debounce.

- `engine_is_built_once_across_debounce_cycles` (`:363`) writes two files in
  separate debounce windows, waits for two update cycles, and asserts the builder
  ran once.
- `changes_under_excluded_directories_do_not_trigger_updates` (`:399`) asserts the
  presence before the absence: it first writes an indexable file and waits for
  exactly one cycle, then writes five files under `target/debug/` and asserts the
  count is still one. Without the presence half, a watcher that never fires would
  pass.
- `index_writes_never_trigger_updates_even_without_default_excludes` (`:444`) does
  the same with `no_default_excludes = true` and writes under `.vera/`, which is
  the configuration that removes the pattern from the matcher.

Three more in `crates/vera-core/src/discovery/mod.rs`, direct tests for
`ExclusionMatcher`, which had none:

- `matcher_agrees_with_discovery_on_a_file_named_like_an_excluded_dir` (`:983`)
  runs `discover_files` on a fixture containing a regular file named `build` and
  a file under `target/debug/`, asserts which of the two discovery actually
  indexes, and then asserts the matcher classifies them the same way. Pinning
  discovery's answer first is what makes the matcher assertion mean anything.
- `matcher_excludes_index_and_vendor_dirs_but_not_sources` (`:1024`) covers
  `.vera`, `.git`, `node_modules`, a source file, and a path outside the root.
- `matcher_applies_extra_excludes` (`:1043`) covers a `--exclude` pattern.

### Reinjection

All four defects reinjected at the head commit; each test fails with the real
production symptom.

`SharedEngine::get` calling the builder every time:

```
thread 'watcher::tests::engine_is_built_once_across_debounce_cycles' panicked at
crates/vera-mcp/src/watcher.rs:388:9:
assertion `left == right` failed: the engine must be built once and reused, not rebuilt per cycle
  left: 2
 right: 1
```

The exclusion filter dropped from the event predicate:

```
thread 'watcher::tests::changes_under_excluded_directories_do_not_trigger_updates' panicked at
crates/vera-mcp/src/watcher.rs:430:9:
assertion `left == right` failed: build output under target/ must not trigger an update cycle
  left: 2
 right: 1
```

The index directory routed through `ExclusionMatcher` instead of its own check:

```
thread 'watcher::tests::index_writes_never_trigger_updates_even_without_default_excludes' panicked at
crates/vera-mcp/src/watcher.rs:479:9:
assertion `left == right` failed: writes into the index directory must never trigger an update cycle
  left: 2
 right: 1
```

`is_excluded` matching every path as a directory:

```
thread 'discovery::tests::matcher_agrees_with_discovery_on_a_file_named_like_an_excluded_dir' panicked at
crates/vera-core/src/discovery/mod.rs:1009:9:
a regular file named `build` is indexed, so it must not be classified as excluded
```

## Verification

Run at the head commit.

| check | result |
|---|---|
| `cargo build --workspace` | clean |
| `cargo fmt --check` | clean |
| `cargo test -p vera-core --lib` | 796 passed, 0 failed |
| `cargo test -p vera-cli --bin vera` | 97 passed, 0 failed |
| `cargo test -p vera-mcp` | 40 passed, 0 failed |
| `cargo clippy -p vera-core --lib` warnings | 5, all pre-existing |

---

---
## Summary by cubic
Reuses a single runtime and embedding provider for the MCP watcher to remove per-debounce rebuilds, aligns event filtering with discovery, and prevents index self-trigger loops. Also rebuilds the engine when the provider identity changes and purges stale chunks when indexed paths move into excluded directories.

- Caches runtime/provider/config/model in a `SharedEngine`, built lazily on first use; the cache key is a provider identity (backend, endpoint, model, query prefix, and embedding runtime options). Runtime is resolved each cycle so identity changes rebuild the engine.
- Filters events via `vera-core`’s `ExclusionMatcher`; probes path type so regular files named like excluded directories are not dropped; `.vera` is always ignored by path.
- Treats a change as relevant if it’s not excluded, is a directory, or is a previously tracked file; this lets updates purge stale chunks when a file becomes an excluded directory.
- Improves `watch()` failure messaging for OS watch limits; serving continues and retries on the next call; concurrency unchanged (cycles remain serialized).
- Tests cover `ExclusionMatcher`, engine rebuild on identity change, watcher behavior with `no_default_excludes`, and purging when an indexed path enters an excluded directory.

<sup>Written for commit 734a45ad2e698d58c9272baacb1a5613e793c511. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/133?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Improvements**
  * File watching now updates indexes more efficiently across repeated changes.
  * Configured exclusions are consistently respected, including when files or directories become excluded.
  * Stale indexed content is automatically removed when files are deleted or converted into directories.
  * Index directories are protected from being indexed accidentally.
  * Watch setup now provides clearer guidance when the operating system’s watch limit is reached.


citron07r:
## Local review (the hosted one was rate limited)

The CodeRabbit check on this PR reported `pass` with the reason **`Review rate limited`**, so nothing was actually reviewed. I ran the review locally from the branch's worktree instead:

```
coderabbit review --committed --base-commit e3d79b3
```

**2 findings, both major, both real.** Each was confirmed by execution before being fixed, not by reading. Both are fixed in e504761.

### 1. Keep `.vera` unconditionally excluded (major, Stability & Availability) — accepted

> `ExclusionMatcher` follows `no_default_excludes`. If that option is true, `.vera` is no longer excluded. An incremental update can then observe its own index writes and start repeated update cycles.

Correct, and reachable two ways rather than one. `load_saved_runtime_config` deserialises the entire `IndexingConfig` out of `~/.vera/config.json`, so both `no_default_excludes: true` **and** a stored `default_excludes` array that omits `.vera` reach the watcher and remove the guard. Routing `.vera` through a config-driven matcher made a load-bearing invariant tunable, which is a regression this branch introduced: the pre-branch code checked `starts_with(repo/.vera)` directly and could not be switched off.

Proven before fixing. The new test with `no_default_excludes = true` fails against the branch as it stood:

```
thread 'watcher::tests::index_writes_never_trigger_updates_even_without_default_excludes' panicked at
crates/vera-mcp/src/watcher.rs:471:9:
assertion `left == right` failed: writes into the index directory must never trigger an update cycle
  left: 4
 right: 1
```

Four cycles from five index writes, each of which would itself write into `.vera` again. Fixed by filtering `index_dir` on its own path, unconditionally, ahead of the matcher (`watcher.rs:165`, `:192`), leaving `ExclusionMatcher` responsible only for the directories indexing skips.

### 2. Use the actual path type for directory-only patterns (major, Functional Correctness) — accepted

> Line 228 passes `true` for every path. A regular file named `target` can match `!target/`, while discovery indexes that file. The watcher then ignores changes to the indexed file.

Correct. `discover_files` applies no extension allowlist, so an extensionless file named `build`, `target` or `dist` at the repo root — an ordinary shell-script name — is indexed, while `is_excluded` classified it as excluded and the watcher dropped every edit to it.

Proven by pinning discovery's answer first, then asserting the matcher agrees. Against the unfixed `is_excluded`:

```
thread 'discovery::tests::matcher_agrees_with_discovery_on_a_file_named_like_an_excluded_dir' panicked at
crates/vera-core/src/discovery/mod.rs:1001:9:
a regular file named `build` is indexed, so it must not be classified as excluded
```

`is_excluded` now calls `path.is_dir()` (`discovery/mod.rs:232`). Two deviations from the suggestion as written, both in the direction it asked for:

- A path that no longer exists is treated as a file, as suggested, so deleting such a file still triggers an update. Deletions *under* `target/` stay excluded regardless, because the parent chain is matched as directories either way — that half of the walk never depended on the probe.
- The one behaviour this gives up is that removing the `target` directory itself now costs one update cycle. That is stated in the description rather than special-cased, because a guard for it would reintroduce the same class of bug for a deleted file named `target`.

### Tests added

`ExclusionMatcher` is a new public API and had no direct tests; integration through the watcher pinned the integration, not the contract. Three added in `crates/vera-core/src/discovery/mod.rs` (`:983`, `:1024`, `:1043`) covering the discovery-agreement case above, the index/vendor directories, a source file, a path outside the root, and an `--exclude` pattern. One added in `crates/vera-mcp/src/watcher.rs` (`:444`) for the `no_default_excludes` case.

All four defects on this branch — including the two the original description already claimed — were reinjected at the head commit and each test fails with the real production symptom. The four failure outputs are quoted in the description.

### Body corrections

Line references re-derived at the head commit. Two claims were corrected rather than restated: `.vera` is no longer described as coming from `IndexingConfig` alongside `target`/`node_modules`, since it is now filtered separately and deliberately; and the description now states the scope boundary it was missing — `ExclusionMatcher` covers `default_excludes` and `--exclude` only, not `.gitignore` or `.veraignore`, so a gitignored directory still costs a no-op cycle. That is the safe direction, but it was not written down.

### Re-verified at e504761

| check | result |
|---|---|
| `cargo build --workspace` | clean |
| `cargo fmt --check` | clean |
| `cargo test -p vera-core --lib` | 796 passed, 0 failed |
| `cargo test -p vera-cli --bin vera` | 97 passed, 0 failed |
| `cargo test -p vera-mcp` | 40 passed, 0 failed |
| `cargo clippy -p vera-core --lib` | 5 warnings, all pre-existing |

PR #146, opened 2026-08-23
fix(retrieval): pick deep-search follow-up symbols deterministically

### LINKED ISSUE #145: fix(retrieval): deep-search follow-up symbols picked from a HashSet make multi-hop results nondeterministic
## Summary

`execute_iterative_search_with_context` in `crates/vera-core/src/retrieval/iterative_search.rs` picks the follow-up symbols for the second hop from a `HashSet`, whose iteration order is randomized per process:

```rust
.collect::<HashSet<_>>()
.into_iter()
.take(5)
```

When the initial results carry more than five distinct symbol names, `.take(5)` selects a process-random subset. Those symbols drive the follow-up searches, so both the result set and merged order differ between two runs of the identical query. The same function feeds RAG fusion's fallback path (`rag_fusion.rs` calls `execute_iterative_search_with_context`), where different follow-up symbols mean genuinely different fused content, not just reordering.

## Reproduction

1. Index a repo where a deep-search query yields more than five distinct symbol names in the initial results.
2. Run `vera deep-search "<query>"` (no completion endpoint configured, so the iterative fallback runs) several times.
3. Observe different follow-up symbols and different final results across runs; retrying to reproduce an earlier result is impossible.

Related: #122 (trailing-period ordering nondeterminism) was fixed at a different site; this one survived it.

## Expected behaviour

Identical input queries must produce identical multi-hop searches. The selection should be deterministic while still preferring relevance - e.g. keep first-appearance order of symbols in the initial (ranked) results and take the first five, or use a sorted structure.

## Suggested fix

Replace the `HashSet` round-trip with a deterministic collection: either `BTreeSet`, or dedupe preserving first-appearance order in the ranked initial results before `.take(5)`. Add a regression test that pins which five symbols are chosen for a >5-symbol fixture.

citron07r:
Closes #145.

## Problem

The follow-up symbols for deep search's second hop were deduped through a `HashSet` and capped with `.take(5)`. Whenever the initial results carried more than five distinct qualifying symbol names, the selection was process-random: two runs of the identical query could issue different second-hop searches and return genuinely different results (not just reordering). This also fed RAG fusion's fallback path, where different symbols mean different fused content.

## Fix

Dedupe while preserving first-appearance order of the ranked initial results (`HashSet` as seen-set only), then take five. Relevance order now decides instead of hash seeds. The selection logic is extracted into a pure `follow_up_symbols_from(&[SearchResult])` so it can be tested without a `SearchContext`.

## Verification

Ran locally on this branch:

- New tests (order under the cap; >5 cap takes first five in ranked order; dedupe + generic-name filtering):
  - **Against the old HashSet round-trip (reverted helper, tests kept):** all three fail (reinjection check).
  - **With the fix:** 3 passed.
- Full module suite: `cargo test -p vera-core --lib iterative_search` — 3 passed, 0 failed.
- `cargo fmt -p vera-core --check`: clean.
- `cargo clippy -p vera-core --lib`: only the 5 pre-existing warnings documented in CLAUDE.md.

Not run: end-to-end multi-hop deep search against a live index (the pure function pins the exact selection contract; the hop loop is unchanged apart from consuming its deterministic output).

---
## Summary by cubic
Make deep-search follow-up symbol selection deterministic. Old behavior deduped via a HashSet and took 5, yielding process-random symbols when >5 qualified; new behavior preserves first-appearance order of ranked results, then takes the first 5. This stabilizes second-hop queries and downstream fusion content.

- Extracts selection into a pure function follow_up_symbols_from that dedupes in order and skips generic names (e.g., main, init), then caps at five; `execute_iterative_search_with_context` now uses it.
- Adds unit tests covering order under the cap, >5 cap behavior, and dedupe plus generic-name filtering.
- Changes confined to `crates/vera-core/src/retrieval/iterative_search.rs`; no API or config changes.

<sup>Written for commit c7aa256c73ec0cf7abcabbd9b626e9ca52f9c9da. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/146?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Improvements**
  * Improved iterative search to identify relevant follow-up symbols in a consistent ranked order.
  * Removed duplicate and overly generic symbol names from follow-up results.
  * Limited follow-up suggestions to a maximum of five items.

### DIFF
--- a/crates/vera-core/src/retrieval/iterative_search.rs
+++ b/crates/vera-core/src/retrieval/iterative_search.rs
@@ -13,6 +13,29 @@ use crate::types::{SearchFilters, SearchResult};
 use super::query_utils::result_key;
 use super::search_service::{SearchContext, SearchTimings};
 
+/// Follow-up symbols for the next hop: deduped, generic noise names skipped,
+/// capped at five, in first-appearance order of the ranked initial results.
+/// A `HashSet` round-trip here used to make the choice process-random whenever
+/// more than five distinct symbols qualified, so two runs of the identical
+/// query could search different second hops and return different results.
+fn follow_up_symbols_from(initial_results: &[SearchResult]) -> Vec<String> {
+    let mut seen = HashSet::new();
+    initial_results
+        .iter()
+        .filter_map(|r| r.symbol_name.clone())
+        .filter(|name| {
+            let lower = name.to_ascii_lowercase();
+            // Skip generic names that would produce noisy results.
+            !matches!(
+                lower.as_str(),
+                "main" | "new" | "default" | "test" | "init" | "run" | "setup"
+            )
+        })
+        .filter(|name| seen.insert(name.clone()))
+        .take(5)
+        .collect()
+}
+
 /// Run an iterative (multi-hop) search.
 ///
 /// 1. Execute the initial query via `context.search`.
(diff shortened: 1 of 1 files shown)


lemon07r:
Superseded by abed08d on master: follow-up symbols are now picked deterministically by ranked first appearance with the same cap of 5 and generic-name filtering, with determinism tests. Closes #145.

Issue #140, opened 2026-08-20
vera-core tests mutate the process environment; convert them to child processes


citron07r:
`vera-core`'s environment-dependent unit tests mutate the process environment. Since #125 they all go through `crate::test_env::EnvVarGuard`, which takes one crate-wide lock and restores what it changed on `Drop`, including while unwinding. That makes them correct against each other, but it does not make `set_var`/`remove_var` sound: a crate-private lock cannot exclude an environment read from library code or from a thread the test did not spawn, and on non-Windows targets `std::env::set_var` is only sound in a single-threaded program. `cargo test` runs the crate's tests as threads of one process.

Raised by CodeRabbit on #125 and agreed to be out of that PR's scope, which was `--lang` value handling and `retrieval.max_rerank_batch`. Filed here so the conversion is tracked rather than lost.

## Current state

Every environment mutation in `vera-core` is in test code and goes through the guard. `grep -rn -E 'env::(set_var|remove_var)|env_lock' crates/vera-core/src/ | grep -v src/test_env.rs` returns nothing.

The call sites, by what they actually test:

| Site | Subject |
|---|---|
| `config.rs:652` | `graph_augmentation_enabled` (`config.rs:300`) |
| `config.rs:695` | `default_max_in_flight_inputs` (`config.rs:186`) |
| `config.rs:750` | `backend_from_env` (`config.rs:310`) |
| `config.rs:760` | `is_local_mode` (`config.rs:289`) |
| `config.rs:823` | `aliases_match_env` (`config.rs:609`) |
| `dynamic_reranker.rs:87` | that `retrieval.max_rerank_batch` beats `VERA_MAX_RERANK_BATCH` |
| `search_service.rs:455`, `search_service.rs:523` | nothing; the variables only make the API provider constructible |
| `test_env.rs:94`, `test_env.rs:138` | the guard itself |

`vera-cli/src/state.rs:384` also mutates the environment, but as production single-threaded CLI startup before any runtime threads exist. Out of scope here.

## What the conversion has to account for

**The first six are tests of environment parsing.** The read is the subject, so replacing it with an injected parameter deletes what they cover, not just how they set it up. They need child processes, not injection. `resolve_backend` already takes `Option<InferenceBackend>`; the two tests pass `None` precisely to reach the environment fallback.

[...]

PR #148, opened 2026-08-23
fix(cli): treat empty and whitespace-only env vars as absent in doctor

### LINKED ISSUE #147: fix(cli): vera doctor scores empty API env vars as present
## Summary

`check_env_group` in `crates/vera-cli/src/commands/doctor.rs` scores API environment variables on *presence* (`std::env::var_os(key).is_some()`), not validity:

```rust
let present = keys.iter().filter(|key| std::env::var_os(key).is_some()).count();
...
n if n == keys.len() => CheckStatus::Ok,
```

Two problems:

1. `EMBEDDING_MODEL_API_KEY=""` (a common shellrc leftover after revoking a key) counts as **present**, so `embedding-api` reports ok. Every consumer treats empty as unset: `setup.rs::read_api_env_var` explicitly trims and filters empties ("Read an env var, treating unset and empty/whitespace-only values the same"), and a provider built with an empty base URL fails at request time. Doctor gives its strongest positive verdict to a config `vera setup` refuses to save and that cannot work.
2. The inverse misleads too: a healthy install configured via `vera setup --api` (which tells the user they can remove the `EMBEDDING_MODEL_*` vars from their shell) shows `warn 0/3 variables present`, because doctor never looks at the saved `config.embedding_api`.

## Reproduction

```bash
EMBEDDING_MODEL_BASE_URL="" EMBEDDING_MODEL_ID="" EMBEDDING_MODEL_API_KEY="" vera doctor
# embedding-api: ok — 3/3 variables present
```

## Expected behaviour

- Empty/whitespace-only values count as absent (matching `read_api_env_var` semantics).
- In API mode, the saved `config.embedding_api` / credentials state should be what doctor reports; raw env presence is only an override signal.

## Suggested fix

Minimum: read with `std::env::var` and treat trim-empty as missing. Better: when the resolved backend is API, check the saved endpoint config and credential file first and report that, falling back to env-var detection for override setups.

citron07r:
Closes #147.

## Problem

`check_env_group` scored API variables with `std::env::var_os(key).is_some()`, i.e. presence only. `EMBEDDING_MODEL_API_KEY=""` (a shellrc leftover after revoking a key) counted as present and earned doctor's strongest positive verdict for a configuration that every consumer treats as unset: `setup::read_api_env_var` explicitly trims and filters empties, and a provider built from an empty base URL fails at request time. The inverse also misled: a healthy `vera setup --api` install showed `warn 0/3 variables present` because doctor never looked at the saved `config.embedding_api`.

## Fix

Presence now requires a non-blank value (`trim()` + non-empty), matching `read_api_env_var` semantics. The decision is extracted into `check_env_group_with` taking an injected getter, so the new tests drive it through a closure instead of mutating process environment variables (keeping clear of #140).

Scope note: this fixes the validity half of #147. Reporting the saved `config.embedding_api` state in API mode is a larger UX change and is deliberately not attempted here; the issue can stay open or be split if that half is still wanted after this lands.

## Verification

Ran locally on this branch:

- New test `empty_env_values_count_as_absent_not_present`:
  - **Without the fix (presence-only logic restored, test kept):** fails — empty values scored present, status Ok instead of Fail (reinjection check).
  - **With the fix:** passes: one valid value out of three reports `Fail`, detail `1/3 variables present`.
- Plus `all_valid_env_values_score_ok` and `no_env_values_at_all_warns` pinning the unchanged Ok/Warn paths.
- `cargo test -p vera-cli --bin vera doctor::`: 6 passed, 0 failed.
- `cargo fmt --check`: clean.
- `cargo clippy -p vera-cli --bin vera`: no new warnings (only pre-existing).

Not run: end-to-end `vera doctor` against a real API backend.

---
## Summary by cubic
Doctor now treats empty and whitespace-only environment variables as absent, aligning with `setup::read_api_env_var`. Previously any set variable (including "") counted as present, so misconfigured API envs could show Ok; now only non-blank values count, which can change verdicts to Warn/Fail.

- Extracts the presence check into `check_env_group_with(getter)` to allow injected lookups in tests.
- Updates env presence logic to `trim()` and require non-empty values; adds tests for empty values, all valid, and none set.
- Does not change reporting of saved `config.embedding_api` in API mode.

<sup>Written for commit 3ef4e4091eaab491b5aeb3b2d7239803eca098e7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API environment checks to correctly identify unset, empty, and whitespace-only values as missing.
  * Added validation coverage for partially configured, fully configured, and unconfigured environments.

### DIFF
--- a/crates/vera-cli/src/commands/doctor.rs
+++ b/crates/vera-cli/src/commands/doctor.rs
@@ -208,9 +208,21 @@ fn failed_check_names(checks: &[DoctorCheck]) -> Vec<&'static str> {
 }
 
 fn check_env_group(name: &'static str, keys: &[&'static str]) -> DoctorCheck {
+    check_env_group_with(name, keys, |key| std::env::var(key).ok())
+}
+
+/// Presence with validity: unset and empty/whitespace-only both count as
+/// absent, matching `setup::read_api_env_var`. An `EMBEDDING_MODEL_API_KEY=""`
+/// leftover in a shellrc used to score as present and earn an ok verdict for a
+/// configuration that cannot work.
+fn check_env_group_with(
+    name: &'static str,
+    keys: &[&'static str],
+    get: impl Fn(&str) -> Option<String>,
+) -> DoctorCheck {
     let present = keys
         .iter()
-        .filter(|key| std::env::var_os(key).is_some())
+        .filter(|key| get(key).is_some_and(|value| !value.trim().is_empty()))
         .count();
 
     let status = match present {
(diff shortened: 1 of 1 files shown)


lemon07r:
Superseded by 592864c (Closes #147): blank/whitespace env values now score as absent in doctor. Review note about saved-config reporting: already covered, main.rs applies saved config into env before dispatch, so doctor reflects the effective config. Closing with thanks.

PR #156, opened 2026-08-23
fix(storage): read commands no longer fabricate an empty metadata.db

### LINKED ISSUE #155: fix(storage): read commands create an empty metadata.db and report success when the index is missing
## Summary

Every read-side guard checks only that the `.vera/` directory exists, never that `metadata.db` exists. `MetadataStore::open` uses `Connection::open`, which **creates** the file when missing, then `init_schema()` runs DDL and WAL pragmas on every open, read or write (`crates/vera-core/src/storage/metadata.rs:152-157`).

So after a crash mid-`store_index` (a window the pipeline's own comments acknowledge) or a deleted/truncated db, every read command fabricates a fresh empty database:

- `vera stats` prints `file_count: 0, chunk_count: 0` and exits 0 — success recorded for an index that does not exist.
- `vera references foo` prints "No callers found for 'foo'." instead of erroring.
- The read command has also written schema + WAL files into `.vera/` unannounced: a repair applied on the read path leaks into disk state.

## Reproduction

```bash
mkdir -p demo && cd demo && mkdir .vera          # directory without metadata.db
vera stats                                        # -> success, all-zero stats
ls .vera                                          # -> metadata.db now exists, created by the READ
```

## Expected behaviour

Read commands must not create database files. A missing `metadata.db` alongside an existing `.vera/` means the index is broken or half-written; the honest answer is an error pointing at re-indexing, not silent zeros.

## Suggested fix

Add `MetadataStore::open_existing` which errors when the file is absent, and use it from the read-side entry points (`stats`, references, structural, regex, bm25/hybrid/exact-match search paths). Keep create-or-open semantics for the indexing pipeline's write path. Optionally tighten the CLI's "no index found" guard to check the db too so users get the friendly hint rather than a deep storage error.

citron07r:
Closes #155.

## Problem

`Connection::open` creates the file and `init_schema` stamps DDL + WAL pragmas on every open, read or write. Every read-side guard only checked that `.vera/` exists — never `metadata.db` — so after a crash mid-`store_index` (or a deleted db), any read command fabricated a fresh empty database: `vera stats` reported all-zero stats with exit 0, `vera references foo` reported "no callers found", and the read left a schema-stamped db behind unannounced. Read-path repair leaking into disk state, plus success recorded for an index that does not exist.

## Fix

- `MetadataStore::open_existing`: errors with a re-index hint when the database file is absent; otherwise delegates to `open`.
- All production read paths switched to it: `stats` (`collect_stats`, `collect_overview_filtered`, symbol lookups), references, structural, regex search, BM25 search, hybrid vector metadata, exact matches, the search-service staleness check, and `vera update`'s pre-read.
- The indexing pipeline's write path keeps create-or-open semantics.
- The CLI's "no index found" guard (`prepare_indexed_repo`) now checks `metadata.db` so users see the friendly hint instead of a storage-layer error.

## Verification

Ran locally on this branch:

- New unit test `open_existing_errors_on_missing_db_without_creating_it`:
  - **Existence check disabled (reinjection):** fails — the open succeeds and fabricates the db.
  - **With the fix:** passes — error names "no index metadata found", and no file is created; also pins that explicit `open` still creates.
- Full vera-core lib suite: **899 passed**, 0 failed.
- Executed end to end: in a directory containing only `.vera/`, `vera stats` exits **1** with "no index metadata found ... run \`vera index <path>\`" and leaves `.vera/` empty. On master it exited **0** with all-zero stats and created `metadata.db`.

Not run: live hybrid/vector searches against a real model backend (the touched call sites are store-open plumbing; behavior beyond the missing-db case is unchanged).

---
## Summary by cubic
Read commands no longer create an empty `metadata.db`. Previously reads created-or-opened and returned empty results with exit 0; now reads open the store read-only, validate all required tables, and error with a re-index hint without creating or modifying files.

- Add `MetadataStore::open_existing`: uses read-only flags and validates the full schema (rejects zero-byte/truncated and partial schemas; error names the first missing table).
- Switch all read paths to it: stats and all retrieval paths (BM25, hybrid, regex, structural, references, type relations, exact matches), search-service checks, and `vera update` pre-read.
- Keep create-or-open on the write path; explicit `open` still creates the database.
- CLI guard now checks for `metadata.db` and shows “run `vera index <path>`” instead of a storage-layer error.
- Verified: unit test ensures failed opens don’t create files and catches partial schemas; end-to-end `vera stats` in a `.vera/`-only directory exits non-zero and leaves the directory untouched.

<sup>Written for commit 78a95f933a86e0fb7f7a0cf37284ab2484f76dff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/156?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Search, statistics, and validation workflows no longer create missing metadata databases automatically.
  - Missing, empty, invalid, or incomplete index metadata is now reported clearly instead of being treated as available.
  - Existing search fallback behavior remains unchanged when metadata cannot be opened.

- **Reliability**
  - Read-only operations now avoid creating or modifying metadata files.
  - Metadata validation ensures required database structures are present before use.
  - Index detection now requires valid metadata files.

### DIFF
--- a/crates/vera-cli/src/commands/update.rs
+++ b/crates/vera-cli/src/commands/update.rs
@@ -59,7 +59,9 @@ pub fn run(path: &str, json_output: bool, options: CommandOptions) -> anyhow::Re
 
     // Check metadata mismatch
     let metadata_path = repo_path.join(".vera").join("metadata.db");
-    if let Ok(metadata_store) = vera_core::storage::metadata::MetadataStore::open(&metadata_path) {
+    if let Ok(metadata_store) =
+        vera_core::storage::metadata::MetadataStore::open_existing(&metadata_path)
+    {
         if let (Some(s_model), Some(s_dim)) = (
             metadata_store.get_index_meta("model_name").unwrap_or(None),
             metadata_store
--- a/crates/vera-cli/src/helpers.rs
+++ b/crates/vera-cli/src/helpers.rs
@@ -174,7 +174,10 @@ pub fn prepare_indexed_repo(
     let cwd = std::env::current_dir()
         .map_err(|e| anyhow::anyhow!("failed to get current directory: {e}"))?;
     let index_dir = vera_core::indexing::index_dir(&cwd);
-    if !index_dir.exists() {
+    // Presence of `.vera/` alone is not validity: after a crash mid-index or a
+    // deleted db the directory exists but holds no metadata. Read commands must
+    // say "no index" rather than fabricate an empty one further down.
+    if !index_dir.join("metadata.db").is_file() {
         anyhow::bail!(
             "no index found in current directory.\n\
              Hint: run `vera index <path>` first to create an index."
--- a/crates/vera-core/src/retrieval/bm25.rs
+++ b/crates/vera-core/src/retrieval/bm25.rs
@@ -36,8 +36,8 @@ pub fn search_bm25(index_dir: &Path, query: &str, limit: usize) -> Result<Vec<Se
     let metadata_path = index_dir.join("metadata.db");
 
     let bm25_index = Bm25Index::open(&bm25_dir).context("failed to open BM25 index for search")?;
-    let metadata_store =
-        MetadataStore::open(&metadata_path).context("failed to open metadata store for search")?;
+    let metadata_store = MetadataStore::open_existing(&metadata_path)
+        .context("failed to open metadata store for search")?;
 
     search_bm25_with_stores(&bm25_index, &metadata_store, query, limit)
 }
--- a/crates/vera-core/src/retrieval/exact_matches.rs
+++ b/crates/vera-core/src/retrieval/exact_matches.rs
@@ -29,7 +29,7 @@ pub(crate) fn augment_exact_match_candidates(
     filters: &SearchFilters,
 ) -> Result<Vec<SearchResult>> {
     let metadata_path = index_dir.join("metadata.db");
-    let Ok(store) = MetadataStore::open(&metadata_path) else {
+    let Ok(store) = MetadataStore::open_existing(&metadata_path) else {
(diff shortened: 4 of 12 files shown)


citron07r:
Fixed in 344f9a6. `open_existing` now opens with `Connection::open_with_flags(SQLITE_OPEN_READ_ONLY)`, skips `init_schema` entirely, and validates that the database has tables before returning: a zero-byte or truncated file fails as "index metadata ... has no schema (the index is incomplete; run `vera index <path>` to rebuild it)" without being modified. The missing-file check stays for the friendlier hint message; read-only open alone would surface a bare SQLite error there.

citron07r:
Follow-up fixed in 78a95f9: `validate_schema` now verifies all six required tables (chunks, file_hashes, file_index_state, index_metadata, references, type_relations) instead of accepting any-tables-present, so partial schemas fail with "incomplete ... run `vera index <path>`" and the error names the first missing table. Regression test added for a chunks-only schema.

lemon07r:
Superseded by 066348f (Closes #155): MetadataStore::open_existing opens SQLITE_OPEN_READ_ONLY and validates all six required tables; stats, call-graph, and update staleness reads fail with a re-index hint instead of fabricating an empty metadata.db. Closing with thanks.

PR #158, opened 2026-08-23
fix(mcp): watcher picks up changes that land mid-update via a trailing cycle

### LINKED ISSUE #157: fix(mcp): watcher drops file events that arrive during a running update
## Summary

The MCP/`vera watch` file watcher (`crates/vera-mcp/src/watcher.rs`) skips a debounced event batch whenever an update is already running:

```rust
if updating_clone.compare_exchange(false, true, Ordering::SeqCst, Ordering::Relaxed).is_err() {
    debug!("Skipping auto-update: previous update still running");
    // "changes will be picked up next cycle"
    return;
}
```

The message promises pickup "next cycle", but there is no next cycle without a further filesystem event: the skipped batch is consumed and discarded. Changes landing while an update runs are lost until something else changes later.

(The update itself does rescan everything changed before it started, since the incremental scan walks the repo — only edits landing *during* the run are affected.)

## Reproduction

1. Start `vera watch .` on an indexed repo (or use the MCP server's auto-watch).
2. Write file A to trigger an update; make the update slow (large repo / slow embedding backend).
3. While "[watch] update complete" has not printed, write file B.
4. After completion, search for content from B: absent from the index, and no further update occurs until another file event arrives.

## Expected behaviour

Changes observed during a running update must be picked up when it finishes.

## Suggested fix

Set a "dirty" flag on the skip path; after `run_incremental_update` resets `updating`, re-run exactly one more update if the flag was set (clear-then-run so edits during the trailing update schedule another).

citron07r:
Closes #157.

## Problem

Event batches arriving while an update was running were skipped and dropped. The skip message promised pickup "next cycle", but no next cycle exists without a further filesystem event: a single edit landing during a long update (large repo, slow embedding backend) left the index stale for that file indefinitely.

## Fix

- The skip path sets a `dirty` flag instead of just returning.
- `run_incremental_update` now loops: after finishing and releasing the updating flag, it re-acquires the lock once more (so a concurrent batch cannot start a second runner in between) and runs again if changes had queued up; edits during the trailing run schedule yet another cycle through the same flag.
- The update body itself is unchanged, extracted as `run_incremental_update_once`.

## Verification

Ran locally on this branch:

- New test `changes_landing_mid_update_are_picked_up_by_a_trailing_cycle`: a 500 ms engine holds the update lock while the second change's debounce window fires into the skip path.
  - **Re-trigger disabled (reinjection):** fails — the update count stays at 1 until timeout.
  - **With the fix:** passes — a trailing cycle runs and the count reaches 2.
- Full watcher module suite: 6 passed. Whole vera-mcp lib: 51 passed, 0 failed.
- `cargo fmt --check`: clean. `cargo clippy -p vera-mcp --all-targets`: no new warnings.

Not run: live `vera watch` against a real model backend (the engine is stubbed; the flag/lock protocol is what changed).

---
## Summary by cubic
Watcher picks up file changes that arrive during an update by running a trailing cycle; previously those batches were dropped and the index stayed stale until another event. Closes races that could leave `dirty` unobserved or cause a redundant extra cycle.

- On skip, set `dirty` and immediately try to acquire `updating`; if won, clear `dirty` and spawn the trailing cycle, eliminating the lost-update window and avoiding double runs.
- `run_incremental_update` loops and re-acquires the lock to run again while `dirty`, preserving a single runner; the update body moved to `run_incremental_update_once`.
- Adds a test that simulates a slow engine and verifies the trailing cycle in `crates/vera-mcp/src/watcher.rs`.

<sup>Written for commit f187438f89dd4d90a8bb0b578c69b801eb16ec57. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/158?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved update monitoring so changes detected during an active update are processed automatically afterward.
  - Prevented changes from being missed while an update is in progress.
  - Added a trailing update cycle when additional changes are detected during processing.

- **Tests**
  - Added coverage verifying that changes made during an update trigger a follow-up update cycle.

### DIFF
--- a/crates/vera-mcp/src/watcher.rs
+++ b/crates/vera-mcp/src/watcher.rs
@@ -280,6 +280,10 @@ fn start_watching_with_runtime(
                     .compare_exchange(false, true, Ordering::SeqCst, Ordering::Relaxed)
                     .is_ok()
                 {
+                    // We own the trailing run; drop the flag we just set so it
+                    // does not schedule a second cycle on top. Events recorded
+                    // from here on re-set it and still get their own cycle.
+                    dirty_clone.store(false, Ordering::SeqCst);
                     let repo = repo_clone.clone();
                     let flag = updating_clone.clone();
                     let dirty_flag = dirty_clone.clone();
--- a/crates/vera-mcp/src/watcher.rs
+++ b/crates/vera-mcp/src/watcher.rs
@@ -261,8 +261,10 @@ fn start_watching_with_runtime(
             // Skip if already updating, but remember that changes arrived:
             // the running update's final scan happened before these events, so
             // dropping the batch here would leave the index stale until some
-            // unrelated future change. The runner re-runs once when it sees
-            // the flag.
+            // unrelated future change. Record the flag first, then try to take
+            // over the trailing run ourselves: if the active runner exited
+            // between our failed CAS and this point, nobody else will observe
+            // the dirty flag, so we must start the cycle here.
             if updating_clone
                 .compare_exchange(false, true, Ordering::SeqCst, Ordering::Relaxed)
                 .is_err()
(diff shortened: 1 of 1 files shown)


citron07r:
Fixed in 09e586a: after storing `dirty` on the skip path, the callback now attempts to acquire the updating lock itself and spawns the trailing cycle when it wins. If the active runner is still running, its own final check observes the flag; if it already exited between the failed CAS and the dirty store, this spawns the run nobody else would have started. Both orderings now converge on a trailing cycle.

lemon07r:
Superseded by 56c76ca on master: mid-update events set a pending flag inside one mutex-guarded UpdateState (the check-and-set is atomic, closing the lost-update race from review), and trailing passes are bounded at MAX_UPDATE_PASSES=8 so continuous edits cannot hot-loop. Closes #157.

PR #184, opened 2026-08-23
fix(storage): neutralize tantivy boolean operators in sanitized queries

### LINKED ISSUE #165: fix(retrieval): sanitize_query leaves tantivy boolean operators live ('NOT NULL' excludes 'null')
# A03 [P2] `sanitize_query` leaves tantivy boolean/exclusion operators live

- **File:** `crates/vera-core/src/storage/bm25.rs:344-355` (used at `bm25.rs:193-197`)
- **Category:** functional / silent wrong results
- **Verified:** yes, against tantivy-query-grammar 0.25.0 source

## Summary

The comment claims the sanitizer strips "characters that the query parser
interprets as operators", but `-`, `+`, `*`, `?` and the operator keywords
`AND` / `OR` / `NOT` / `IN` pass through untouched:

```rust
fn sanitize_query(query_text: &str) -> String {
    query_text.chars().map(|c| match c {
        ':' | '(' | ')' | '[' | ']' | '{' | '}' | '^' | '~' | '!' | '\'' | '"' | '`' => ' ',
        _ => c,
    }).collect()
}
```

Verified grammar behaviour: `-term` parses as `MustNot`, `+term` as `Must`,
a bare `*` becomes `UserInputLeaf::All` (match-all-documents), and
`NOT <leaf>` negates.

## Failure scenario

- Query `"NOT NULL constraint"` (uppercase, extremely common in SQL-related
  code search) parses as `MustNot("null") AND "constraint"`: every chunk
  mentioning "null" is silently excluded from BM25 results.
- Any query containing a space-delimited `-term` excludes that term; the user
  asked for matching and got exclusion.
- A bare `*` token returns arbitrary chunks.

All of these degrade results invisibly: no error, no signal. Hyphenated words
like `e-mail` are safe (the occur symbol only applies at leaf start).

## Steps to reproduce

1. Index a repo containing the string "NOT NULL constraint".
2. `vera search "NOT NULL constraint"` and compare against
   `vera search "null constraint"`.
3. The first query returns strictly fewer / different hits because "null"
   chunks were excluded.

## Suggested fix

Also blank `-`, `+`, `*`, `?`, and neutralize standalone operator words
(lowercase them, which the grammar treats as ordinary terms) or escape them
via the parser's `\` escaping. Add a regression test pinning that
`search("NOT NULL")` matches the same chunk set as `search("not null")`.

citron07r:
Closes #165.

## Problem

`sanitize_query` stripped punctuation operators but left tantivy's word-level operators live: `-term` parsed as MustNot (so `"NOT NULL constraint"` silently excluded every chunk mentioning "null"), `+term` as Must, a bare `*` as match-all-documents, and uppercase `AND`/`OR`/`NOT` as boolean operators. All of these degraded results invisibly.

## Fix

After the existing char pass, each whitespace-separated token is normalized:

- leading `-`/`+` stripped at word boundaries only — hyphenated terms like `e-mail` survive;
- a bare `*` is dropped; attached wildcards like `auth*` stay;
- uppercase `AND`/`OR`/`NOT` are lowercased into ordinary terms, which is what the user typed.

## Verification

Ran locally on this branch:

- `sanitized_query_does_not_exclude_on_uppercase_not`: searches `"NOT NULL missing"` vs `"NULL missing"` through a real in-memory Tantivy index and asserts identical hit sets.
  - The fixture gained a doc containing NULL — **without it the test passed vacuously against the old code** (nothing to exclude), which reinjection exposed. Existing doc-count assertions updated for the new fixture entry.
  - **Old implementation (reinjected):** both new tests fail.
  - **With fix:** all pass.
- `sanitize_query_neutralizes_word_operators_but_keeps_hyphenated_terms` pins each rule (`-term +other`, `e-mail`, bare vs attached `*`, operator words).
- Full vera-core suite: **900 passed**, 0 failed.
- `cargo fmt --check`: clean. `cargo clippy -p vera-core --lib`: no new warnings.

Not run: end-to-end ranking-quality comparison on a large index (the sanitizer's contract is pinned at the unit/integration level above).

---
## Summary by cubic
Neutralizes Tantivy boolean operators during query sanitization to prevent invisible filtering and match-all queries. Previously `-term`, `+term`, bare `*`, and uppercase `AND/OR/NOT` were applied as operators; now we strip leading +/- at word start, drop bare `*`, and lowercase operator words so they index as ordinary terms.

- The new token-level normalization runs after the existing char-level pass in `sanitize_query` (`bm25.rs`); hyphenated terms (e.g., e-mail) and attached wildcards (e.g., auth*) are preserved.
- Tests: fixture gains a NULL-containing doc; doc-count assertions updated (8 → 9). New tests verify that "NOT NULL ..." matches the same set as "NULL ..." and pin each sanitization rule.

<sup>Written for commit d8bf4814aaf18c7e36f985291f1ebfe6a2b260fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/184?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search query handling for boolean operators and wildcard characters.
  * Preserved hyphenated search terms during query sanitization.
  * Added support for safely handling documents containing NULL values.
  * Improved recognition of uppercase boolean operators such as `NOT`.

* **Tests**
  * Expanded search coverage for operator neutralization, wildcard handling, hyphenated terms, and NULL-containing documents.

### DIFF
--- a/crates/vera-core/src/storage/bm25.rs
+++ b/crates/vera-core/src/storage/bm25.rs
@@ -342,13 +342,35 @@ fn build_schema() -> Bm25Schema {
 }
 
 fn sanitize_query(query_text: &str) -> String {
-    query_text
+    let cleaned: String = query_text
         .chars()
         .map(|c| match c {
             ':' | '(' | ')' | '[' | ']' | '{' | '}' | '^' | '~' | '!' | '\'' | '"' | '`' => ' ',
             _ => c,
         })
-        .collect()
+        .collect();
+
+    // The char pass above cannot neutralize the word-level operators: tantivy
+    // parses a leading `-term` as MustNot (so "NOT NULL constraint"-style
+    // queries silently excluded "null"), a bare `*` as match-all-documents,
+    // and uppercase AND/OR/NOT as boolean operators. Word-boundary stripping
+    // keeps hyphenated terms like "e-mail" intact; lowercasing the operator
+    // words turns them into ordinary terms, which is what the user typed.
+    cleaned
+        .split_whitespace()
+        .map(|word| {
+            let stripped = word.trim_start_matches(['-', '+']);
+            if stripped == "*" {
+                return String::new();
+            }
+            match stripped {
+                "AND" | "OR" | "NOT" => stripped.to_ascii_lowercase(),
+                _ => stripped.to_string(),
+            }
+        })
+        .filter(|word| !word.is_empty())
+        .collect::<Vec<_>>()
+        .join(" ")
 }
 
 fn looks_path_weighted(query_text: &str) -> bool {
@@ -384,6 +406,13 @@ mod tests {
                 file_path: "src/main.rs",
                 content: "struct Config {\n    name: String,\n    port: u16,\n}",
                 symbol_name: Some("Config"),
+                language: "rust",
+            },
+            Bm25Document {
+                chunk_id: "src/db.rs:0",
+                file_path: "src/db.rs",
+                content: "fn save(row: Row) {\n    if row.id is NULL {\n        bail!(\"missing id\");\n    }\n}",
+                symbol_name: Some("save"),
                 language: "rust",
             },
             Bm25Document {
@@ -435,7 +464,7 @@ mod tests {
     fn insert_and_count() {
         let index = Bm25Index::open_in_memory().unwrap();
         index.insert_batch(&sample_docs()).unwrap();
-        assert_eq!(index.doc_count().unwrap(), 8);
+        assert_eq!(index.doc_count().unwrap(), 9);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


lemon07r:
Superseded by abed08d on master. The landed sanitize_query neutralizes leading -/+, edge wildcards, and standalone uppercase AND/OR/NOT/IN, and unlike both earlier drafts it preserves hyphenated identifiers (e-mail stays one term). The operator test compares the actual chunk_id result sets, addressing the review note about length-only comparison. Closes #165.

PR #185, opened 2026-08-23
fix(embedding): apply document prefixes on the API embedding path

### LINKED ISSUE #166: fix(embedding): API path never prefixes documents even when it detects the query prefix
# A09 [P2] API embedding path never prefixes documents, even when it auto-detects the query prefix

- **File:** `crates/vera-core/src/embedding/provider.rs:458-477` (impl), `provider.rs:196-215` (`default_query_prefix_for_model`), hook contract at `provider.rs:83-88`, indexing funnel at `provider.rs:1155`
- **Category:** functional / retrieval quality

## Summary

`default_query_prefix_for_model` recognizes the e5 family (`"query: "`) and BGE,
but there is no `prepare_document_text` override anywhere on the API side:

```rust
impl EmbeddingProvider for OpenAiProvider {
    fn prepare_query_text(&self, query: &str) -> String { /* prefix applied */ }
    // no prepare_document_text anywhere
}
```

The trait hook exists precisely for this ("Applied on the indexing path only"),
and the indexing funnel calls it (`document_prefix_overhead`), so the asymmetry
is one-sided: local models get both prefixes, API models get queries only.

## Failure scenario

`EMBEDDING_MODEL_ID=intfloat/e5-large-v2` against a TEI/vLLM gateway. Queries
are embedded as `"query: ..."`, indexed passages as bare text, but e5's contract
is `query:` / `passage:` pairs. The index degrades silently; nothing errors and
no staleness guard notices because the vectors still "work".

## Steps to reproduce

1. `EMBEDDING_MODEL_BASE_URL=<TEI> EMBEDDING_MODEL_ID=intfloat/e5-large-v2 vera index .`
2. Capture the request bodies (proxy or server log): documents carry no
   `passage: ` prefix while search requests carry `query: `.
3. Compare ranking quality against a local e5 run with both prefixes.

## Suggested fix

Mirror the local config: give `EmbeddingProviderConfig` an optional
`document_prefix` populated alongside the query prefix (`e5 -> "passage: "`),
and override `prepare_document_text` on `OpenAiProvider`. The
`document_prefix_overhead` probe already makes the byte budget correct for free.

citron07r:
Closes #166.

## Problem

`default_query_prefix_for_model` auto-detected query prefixes for e5/BGE/Qwen3/CodeRankEmbed, but nothing overrode `prepare_document_text` on the API side. Queries embedded as `"query: ..."` while indexed passages went bare — breaking e5's asymmetric `query:`/`passage:` contract silently over TEI/vLLM gateways. The local path already applies both sides; the API path was one-sided despite the trait hook existing precisely for this and the indexing funnel calling it (`document_prefix_overhead`).

## Fix

- `EmbeddingProviderConfig` gains `document_prefix: Option<String>`, read from a new `EMBEDDING_DOCUMENT_PREFIX` env override.
- Auto-detected alongside the query prefix: e5 → `"passage: "`; instruction-style families (Qwen3, CodeRankEmbed, BGE's "Represent this sentence...") keep bare passages since their query prefix is an instruction block, not half of a pair.
- `OpenAiProvider::prepare_document_text` applies it on the indexing path.
- The existing `document_prefix_overhead` byte-budget probe reads the trait method, so chunk budgets adjust automatically.

## Verification

Ran locally on this branch:

- `api_document_prefix_applied_on_indexing_path`: e5-configured provider must produce `"query: find foo"` for queries and `"passage: fn foo() {}"` for documents.
  - **Override removed (reinjection):** fails — documents come back unprefixed.
  - **With fix:** passes.
- `e5_model_detects_both_prefixes`: pins detection (e5 gets both; Qwen3/unknown get document-side None).
- Full vera-core suite: **902 passed**, 0 failed.
- `cargo fmt --check`: clean. `cargo clippy -p vera-core --lib`: no new warnings.

Not run: live embedding against a TEI endpoint comparing retrieval quality before/after.

---
## Summary by cubic
Applies passage prefixes for asymmetric models on the API path and neutralizes `tantivy` operators in BM25 query sanitization, fixing retrieval mismatches and accidental exclusions.

- Old: API prefixed queries (e.g., "query: …") but embedded documents bare; terms like "NOT NULL" were parsed as boolean and excluded matches. New: API applies a document prefix (e.g., "passage: …") when needed and sanitization strips word-level operators, so searches treat "NOT/AND/OR" as terms and ignore bare "*". Side effects: chunk budgets account for document prefixes; search results no longer drop documents due to operator parsing.

**Embedding (API path)**
- Adds `document_prefix` to `EmbeddingProviderConfig`, set via `EMBEDDING_DOCUMENT_PREFIX` or auto-detected (e5 → "passage: "; instruction-style families keep bare passages).
- `OpenAiProvider::prepare_document_text` applies the prefix on indexing, matching the local path.
- Migration: if you used e5 over the API, reindex documents to align embeddings with "passage: " prefixes.

**BM25 query sanitization**
- `sanitize_query` now:
  - Strips leading +/- at word boundaries, drops bare "*", and lowercases AND/OR/NOT into ordinary terms while preserving hyphenated words (e.g., "e-mail").
- Test fixture adds a "NULL" sample; doc count assertions increase by one. No rollout action required.

<sup>Written for commit 58e0bde3afae6953d02d5da17de99363a4f516ea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added optional document prefixes for compatible embedding providers.
  * Automatically applies the recommended `passage:` prefix for E5 embedding models.
  * Improved batch processing to account for prefix space.

* **Bug Fixes**
  * Improved search query handling for leading operators, wildcard terms, and Boolean keywords.
  * Preserved hyphenated terms and attached wildcards during query processing.
  * Ensured queries containing `NOT` behave as expected.

### DIFF
--- a/crates/vera-core/src/embedding/provider.rs
+++ b/crates/vera-core/src/embedding/provider.rs
@@ -126,6 +126,11 @@ pub struct EmbeddingProviderConfig {
     /// Optional prefix prepended to query text for asymmetric embedding models.
     /// Read from `EMBEDDING_QUERY_PREFIX` env var.
     pub query_prefix: Option<String>,
+    /// Optional prefix prepended to indexed passage text, the other half of an
+    /// asymmetric contract (e.g. e5's "passage: "). Read from
+    /// `EMBEDDING_DOCUMENT_PREFIX` env var; auto-detected alongside the query
+    /// prefix so the API path matches the local path's two-sided behavior.
+    pub document_prefix: Option<String>,
 }
 
 impl std::fmt::Debug for EmbeddingProviderConfig {
@@ -150,6 +155,7 @@ impl EmbeddingProviderConfig {
             timeout: Duration::from_secs(30),
             max_retries: 3,
             query_prefix: None,
+            document_prefix: None,
         }
     }
 
@@ -170,9 +176,14 @@ impl EmbeddingProviderConfig {
             .ok()
             .filter(|s| !s.is_empty())
             .or_else(|| default_query_prefix_for_model(&model_id));
+        let document_prefix = std::env::var("EMBEDDING_DOCUMENT_PREFIX")
+            .ok()
+            .filter(|s| !s.is_empty())
+            .or_else(|| default_document_prefix_for_model(&model_id));
 
         let mut config = Self::new(base_url, model_id, api_key);
         config.query_prefix = query_prefix;
+        config.document_prefix = document_prefix;
         Ok(config)
     }
 
@@ -210,6 +221,20 @@ fn default_query_prefix_for_model(model_id: &str) -> Option<String> {
     } else {
         // Unrecognized model: try fetching prefix from HuggingFace.
         fetch_query_prefix_from_hf(model_id)
+    }
+}
+
+/// Auto-detect the passage-side prefix matching [`default_query_prefix_for_model`].
+///
+/// Only families with a genuinely asymmetric query/passage contract get one;
+/// a model whose query prefix is an instruction block (Qwen3, CodeRankEmbed,
+/// BGE's "Represent this sentence...") embeds passages bare.
+fn default_document_prefix_for_model(model_id: &str) -> Option<String> {
+    let id = model_id.to_lowercase();
+    if id.contains("e5-") || id.contains("e5_") {
+        Some("passage: ".into())
+    } else {
+        None
     }
 }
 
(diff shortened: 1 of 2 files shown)


lemon07r:
Superseded by 6b291b9 on master, which also resolves the P1 review finding: the document prefix identity is now stored in index metadata, search falls back to BM25-only with a warning on mismatch, and update refuses to append to a mismatched vector space. On the bm25 review note: dropping an explicit -exclusion is deliberate; sanitize_query neutralizes operators rather than silently changing semantics. Closes #166.

PR #186, opened 2026-08-23
fix(config): parse free VRAM from rocm-smi rows that start with GPU

### LINKED ISSUE #167: fix(config): ROCm VRAM detection discards every rocm-smi data row
# A10 [P2] ROCm VRAM detection filters out every data row, then would read the wrong column anyway

- **File:** `crates/vera-core/src/config.rs:498-531` (`detect_rocm_gpu_info`)
- **Category:** functional / platform-specific silent degradation

## Summary

`rocm-smi --showmeminfo vram --csv` emits rows shaped:

```
GPU[0],vram total used memory (bytes),<used>,vram total free memory (bytes),<free>
```

The parser skips every row starting with `"GPU"`:

```rust
for line in stdout.lines().skip(1) {
    let trimmed = line.trim();
    if trimmed.is_empty() || trimmed.starts_with("GPU") {
        continue;                       // <- discards 100% of data rows
    }
    ...
    vram_free_mb = line.split(',').filter_map(...).next()...  // column 3 = USED bytes, not free
```

So `vram_free_mb` is always `None`, the fingerprint always falls back to
`host_fingerprint`. And even without the filter, `.next()` grabs the *used*-bytes
column rather than free.

## Failure scenario

An `onnx-jina-rocm` user with a 16 GB card hits the "Could not detect VRAM"
branch in `adjust_for_backend` (`config.rs:388-395`): batch size pinned to 16
regardless of hardware, and the batch-scaler persistence profile keyed on the
generic host string instead of the device.

## Steps to reproduce

1. On any ROCm machine run `rocm-smi --showmeminfo vram --csv` and observe every
   data row starts with `GPU[0]`.
2. Run `VERA_LOG=debug vera ... --onnx-jina-rocm` and observe the VRAM-detection
   fallback path taken every time.

## Suggested fix

Strip the `GPU[n]` prefix before parsing and select the value following the
`free memory` label (or use `--showmeminfo vram --json` where supported). Add a
unit test over a frozen sample of real rocm-smi output.

citron07r:
Closes #167.

## Problem

Every data row of `rocm-smi --showproductname --showmeminfo vram --csv` starts with `GPU[0]`, so the `starts_with("GPU")` guard skipped 100% of them: `vram_free_mb` was always `None`, ROCm installs got the generic host fingerprint, and `adjust_for_backend` pinned batch size to 16 regardless of hardware. And even without the filter, `.next()` grabbed the first numeric column — the *used*-bytes field — not free.

## Fix

New pure parser `parse_rocm_free_memory_bytes(line)`:

- matches rows containing the `free memory` label (so used-memory rows are rejected),
- reads the value following the label — both the CSV form (`...,vram total free memory (bytes),456`) and the colon form (`vram total free memory (bytes): 456`),
- the loop no longer skips `GPU`-prefixed lines; the fingerprint still prefers non-GPU, non-memory rows (the product name).

## Verification

Ran locally on this branch:

- Two new tests over frozen sample output shapes:
  - `rocm_free_memory_parses_label_value_rows`: single-row label/value CSV → correct byte count; colon form; used-memory row rejected.
  - `rocm_fingerprint_ignores_gpu_header_lines`: product row wins for fingerprint, GPU-header rows never do, free bytes parsed.
- **Reinjection:** restoring the old `starts_with("GPU")` skip makes `rocm_fingerprint_ignores_gpu_header_lines` fail (`None` instead of parsed value).
- Full vera-core suite: **900 passed**, 0 failed. `cargo fmt --check`: clean. Clippy: no new warnings.

Not run: against a physical ROCm machine (samples are frozen shapes of documented rocm-smi output; if a maintainer can run `rocm-smi --showproductname --showmeminfo vram --csv` on hardware, the parser handles that exact shape).

---
## Summary by cubic
Fixes ROCm VRAM detection and fingerprinting by parsing the “vram total free memory (bytes)” value and accepting `GPU[0]`-prefixed product rows. Previously we skipped or rejected these rows, so free VRAM was None, the ROCm fingerprint fell back to host, and batch size stayed conservative.

- Adds `parse_rocm_free_memory_bytes` to match the free-memory label and read the following value in both CSV and “label: value” forms; caller converts bytes to MB.
- Extracts the loop as `parse_rocm_gpu_info`; treats only the bare CSV header field “GPU” as a header (removes the `contains("gpu")` rejection), ignores memory tables, and uses `GPU[0]`-prefixed product rows for the fingerprint.
- Tests lock in parsing and fingerprint behavior; restoring the old GPU-row skip or the `contains("gpu")` filter reproduces the bug.
- Scope: ROCm detection only. No config or API changes. Optional check: run `rocm-smi --showproductname --showmeminfo vram --csv` and confirm free VRAM is reported.

<sup>Written for commit 8dd51fe411467515326a07c65153634d38da7378. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VeraTools/Vera/pull/186?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ROCm GPU detection across all available entries.
  * GPU fingerprints now exclude headers and use more reliable identifying data.
  * Free-memory readings are parsed correctly from labeled values, including inline and colon-separated formats.
  * Used-memory readings are no longer mistaken for available memory.

### DIFF
--- a/crates/vera-core/src/config.rs
+++ b/crates/vera-core/src/config.rs
@@ -510,21 +510,21 @@ fn detect_rocm_gpu_info() -> GpuInfo {
     let stdout = String::from_utf8_lossy(&output.stdout);
     let mut vram_free_mb = None;
     let mut fingerprint_line = None;
-    for line in stdout.lines().skip(1) {
+    for line in stdout.lines() {
         let trimmed = line.trim();
-        if trimmed.is_empty() || trimmed.starts_with("GPU") {
+        if trimmed.is_empty() {
             continue;
         }
-        if fingerprint_line.is_none() {
+        // Product row: not a GPU index header and not a memory table.
+        let lower = trimmed.to_ascii_lowercase();
+        if fingerprint_line.is_none() && !lower.contains("gpu") && !lower.contains("memory") {
             fingerprint_line = Some(trimmed.replace(", ", "|").replace(',', "|"));
         }
         if vram_free_mb.is_none() {
-            // rocm-smi reports bytes; convert to MB.
-            vram_free_mb = line
-                .split(',')
-                .filter_map(|s| s.trim().parse::<u64>().ok())
-                .next()
-                .map(|bytes| bytes / (1024 * 1024));
+            if let Some(bytes) = parse_rocm_free_memory_bytes(trimmed) {
+                // rocm-smi reports bytes; convert to MB.
+                vram_free_mb = Some(bytes / (1024 * 1024));
+            }
         }
     }
     GpuInfo {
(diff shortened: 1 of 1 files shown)


citron07r:
Fixed in ddc2a0e..854aca4: only the bare CSV header field `GPU` is rejected now; `GPU[0],Card series,...` product rows set the fingerprint. The loop moved into a pure `parse_rocm_gpu_info(stdout)` and the test drives that directly — my first attempt had the test re-implementing the loop inline, which made the regression check vacuous (caught on reinjection).

citron07r:
Fixed: only the bare CSV header field `GPU` is treated as a header now; `GPU[0],Card series,...` product rows set the fingerprint. The parsing loop moved into a pure `parse_rocm_gpu_info(stdout)` and the regression test drives that directly (an earlier attempt had the test copy the loop inline, which made it vacuous — caught on reinjection).

lemon07r:
Superseded by 6b291b9 on master with a reworked parser covering the review themes: both rocm-smi CSV layouts (row-local labels and header columns, GPU[n] and cardN rows), free VRAM derived as total-used when no free column exists, fingerprints built only from stable non-memory fields so live VRAM numbers cannot drift the batch-scaler profile key, u64 parsing (negatives rejected), and a defined near-zero behavior (batch 1, memory limit floored at 128MB instead of decaying to unlimited). Closes #167.

